PR #6, opened 2017-02-02
Fix GFM table issue

VadimBrodsky:
The pipe in the sample Regex was breaking the Makrdown table on the [Webpack docs page](https://webpack.js.org/plugins/uglifyjs-webpack-plugin/).
Solution is based on this -- https://github.com/gitlabhq/gitlabhq/issues/1238


bebraw:
Nice one. Thanks! 👍

PR #9, opened 2017-02-15
update plugin

### LINKED ISSUE #8: Update plugin
[Several commits](https://github.com/webpack/webpack/commits/master/lib/optimize/UglifyJsPlugin.js) have been applied to the plugin in webpack/webpack, so we should update it here as well.

Simply copying the new version will lead to a bunch of linting errors, because the projects use different eslint configurations.

SebastianS90:
I just copied the webpack/webpack lib/optimize/UglifyJsPlugin.js and fixed the [relative requires](https://github.com/webpack/webpack/blob/d0f1f6e77009599618b571f9d57bb2a53ba1a583/lib/optimize/UglifyJsPlugin.js#L11:L12).

TODO:
- [x] Update Changelog (look through commits in webpack/webpack)
- ~~Update eslint configuration to be the same as in webpack/webpack~~
- ~~Update remaining js files to be conform with the new eslint configuration~~

(btw, is the pre-push hook really necessary? It prevents me from pushing even to my own fork)

closes #8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+0.1.6 / 2017-02-15
+==================
+
+  * Allow to extract comments to separate file [webpack/webpack#4134](https://github.com/webpack/webpack/pull/4134)
+  * Allow to supress uglifyjs warnings [webpack/webpack#4200](https://github.com/webpack/webpack/pull/4200)
+  * Use const where applicable [1d78f99](https://github.com/webpack/webpack/commit/1d78f990a6af1a5b85e45b345b4f7861b03bb12b)
+  * Avoid reassigning stream with different value [92364dc](https://github.com/webpack/webpack/commit/92364dc0f72ad8c3ac0bdc74e46766658086d83a)
+  * Avoid reassigning files with different value [104398f](https://github.com/webpack/webpack/commit/104398f87a707a601427e666ed318d2338e8a003)
+
 0.1.5 / 2017-02-15
 ==================
 
--- a/src/index.js
+++ b/src/index.js
@@ -1,152 +1,236 @@
 /*
-  MIT License http://www.opensource.org/licenses/mit-license.php
-  Author Tobias Koppers @sokra
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
 */
+"use strict";
 
-
-const SourceMapConsumer = require('source-map').SourceMapConsumer;
-const SourceMapSource = require('webpack-sources').SourceMapSource;
-const RawSource = require('webpack-sources').RawSource;
-const RequestShortener = require('webpack/lib/RequestShortener');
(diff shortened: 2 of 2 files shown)


bebraw:
Can you update the change log?

How does pre-push fail for you? The pre-push thinger gained a Windows fix a while ago that might help. I can also change it to other solution.

bebraw:
I see now, pre-push probably failed with lint. `eslint --fix` should fix most of those.

SebastianS90:
I'll work on this a bit (see TODO List in updated post above)
Can you give me a pointer to the `webpack-defaults` that you mentioned [here](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/8#issuecomment-280059291)?

bebraw:
Sure. See https://github.com/webpack-contrib/webpack-defaults . The idea is that the command will sync to defaults maintained in that repo. Still a bit experimental.

bebraw:
You can skip the pre-push hook with `git push --no-verify`.

SebastianS90:
Okay, after looking into `webpack-defaults` I can summarize:
* it pulls in the eslint configuration from [eslint-config-webpack](https://github.com/webpack-contrib/eslint-config-webpack)
* but that project uses a different configuration than webpack/webpack
* the project is still unreleased
* correct me if I'm wrong, but I guess the ultimate goal is to have webpack/webpack using this eslint configuration, too

So what I did instead of trying to get the same eslint configuration:
* make the plugin pass eslint as configured in this repository
* added the changelog entries

What remains as TODO for the future:
* Make sure this repository uses the same eslint configuration as webpack/webpack, preferably a configuration provided in `eslint-config-webpack` that can be simply imported

bebraw:
> correct me if I'm wrong, but I guess the ultimate goal is to have webpack/webpack using this eslint configuration, too

Yeah, I'll poke about that. Thanks for the insights.

I'll try to get some answers and merge the PR. 👍

bebraw:
I have a decision on this. Webpack core and contrib will follow styles of their own. That said, to keep the core and this plugin in sync, we should move this plugin to follow the core `.eslintrc`. After that these operations would be just copying (minimal effort).

How does that sound?

SebastianS90:
Sounds good! Unfortunately, due to time constraints, I cannot do the refactoring.

You could merge this PR and afterwards revert 71c3d3f6ab688518971088554ef9f68982af1f0c. Then you have an almost-copy of the plugin from webpack/webpack, where only the [two relative requires](https://github.com/webpack/webpack/blob/d0f1f6e77009599618b571f9d57bb2a53ba1a583/lib/optimize/UglifyJsPlugin.js#L11:L12) have been adjusted. Afterwards you need to adjust `.eslintrc` and the other javascript files such that the linter does not complain.

bebraw:
Ok, sounds good. Thanks. 👍

PR #25, opened 2017-04-12
Fix the UglifyJS2 README link

Dru89:
Markdown doesn't support having markdown inside of HTML, so the UglifyJS link wasn't rendering properly.  Now it will.


bebraw:
Thanks. 👍

PR #43, opened 2017-06-19
docs(README): fix description rendering (webpack.js.org)

stof:
webpack.js.org removes all content until the first `h1`. this leaves the `<p>` tag indented and so becoming a code block:
webpack_plugin_doc

Here, I moved the link to markdown below the centered div.
An alternative could had been to keep it inside the div, but dropping the indentation.


README.md
```diff
@@ -12,9 +12,10 @@
       src="https://cdn.rawgit.com/webpack/media/e7485eb2/logo/icon.svg">
   </a>
   <h1>UglifyJS Webpack Plugin</h1>
-	<p>This plugin uses <a href="https://github.com/mishoo/UglifyJS2/tree/v2.x">UglifyJS v2</a> to minify your JavaScript.<p>
```
michael-ciniawsky: The paragraph closing tag seems to be just off :) `<p>`

stof: this would not fix the issue. The issue is that the markdown source becomes this after the stripping done by the script:

```markdown
	<p>This plugin uses <a href="https://github.com/mishoo/UglifyJS2/tree/v2.x">UglifyJS v2</a> to minify your JavaScript.<p>
</div>

> Note that webpack contains the same plugin under `webpack.optimize.UglifyJsPlugin`. This is a standalone version for those that want to control the version of UglifyJS. The documentation is valid apart from the installation instructions in that case.
```

And the indentation before the `<p>` tag makes it a codeblock.

michael-ciniawsky:
```
<p>This plugin uses <a href="https://github.com/mishoo/UglifyJS2/tree/v2.x">UglifyJS v2</a> to minify your JavaScript. --><p><-- // </p>
```

Why does it become a code block 🙃 ? The markdown parser should ignore `html` by default?

stof: The issue is that this is not an HTML block anymore (because the unindented HTML before it gets deleted by the script fetching the readme). It is now a YAML codeblock, showcasing some HTML markup in the codeblock.
If you want that this stays a HTML block when rendering the website, you must ensure that the first line of HTML *in the rendered document* is unindented. And the webpack.js.org website removes the beginning of the readme (until the first `<h1>` it finds, so this `<p>` was becoming the first HTML line (but it is indented)

michael-ciniawsky: hmmm.. I see 

cc @bebraw @skipjack How should we handle this, update the markdown parsing logic or move descriptions under the header?

bebraw: The real solution would be to figure out a nice way at **webpack-defaults** and then apply that across projects. There was some talk about using **markdown-magic** and templating.

stof: hmm indeed. All projects created using webpack-defaults will suffer from this, as the `<p>` tag is part of the template. I will let you find a solution then.

michael-ciniawsky: Yep we need to specify it in `webpack-defaults`aswell then, but the question is still how we should do it :). Would a  e.g `.trim()` somewhere in webpack/webpack.js.org bee enough to avoid the parsing error (as `code` block) for now? If I'm not mistaken most READMEs use `<p>${Description}</p>` atm 😟, so this could be a lot of chore if unresolveable logic wise 😥

skipjack: Yeah idk I'm not exactly sure why the indentation is causing issues but if that's the case, maybe the markdown parsing process or readme parsing process should be tweaked for now. I agree with Juho that it would be really nice to agree on some standards at __webpack-defaults__ because I feel like a lot is falling on our markdown parsing process in the docs repo. 

I'm not too familiar with __markdown-magic__ but I feel like one solution would be to just support the exact same markdown extensions that github does (which is basically what we're hackily trying to do right now). However, they haven't open-sourced their parser as far as I know although I do think I've read that they use `marked` and customize it (as do we). I was hoping to use [marksy](https://github.com/cerebral/marksy) to clean things up as I think it'll allow cleaner customizations but that also uses `marked` under the hood and doesn't support everything github does 😢 .

TLDR; yeah I think we'd accept a hack for now on the docs repo (either in the markdown or readme processing) but, as Juho said, we really need to come to some agreement as to what should be supported and how because the problem will just keep getting worse.

michael-ciniawsky:
Needs to be handled by [webpack-defaults #73](https://github.com/webpack-contrib/webpack-defaults/issues/73) Thx :)

Issue #49, opened 2017-06-22
can not support es6 minification after upgrade to 0.4.6


lili21:
the `0.4.6` version move the `uglify2` to `dependencies`, so I can't specify the `uglify2` version through `yarn add git://github.com/mishoo/UglifyJS2#harmony-v2.8.22 --dev`. cause the plugin will use its own dependency - `uglify-js^2.8.29`.

pkeuter:
Duplicate of #32. See https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/32#issuecomment-300421227 for a workaround.

paulpflug:
this is no duplicate.. the workaround doesn't work anymore

Hotell:
Yup this is very unfortunate. Patch version breaking stuff... :( https://travis-ci.org/Hotell/typescript-lib-starter/builds/244709499?utm_source=github_status&utm_medium=notification

pkeuter:
It is a duplicate. The workaround doesn't work exactly as noted there anymore, but if you change `uglify-js "^2.8.29"` to `uglify-js "^2.8.22"`. Everything is fine. Still an ugly workaround though.

lili21:
I don't get it.

pkeuter:
```
uglify-js@^2.6, uglify-js@^2.8.22, "uglify-js@git://github.com/mishoo/UglifyJS2#harmony-v2.8.22":
  version "2.8.22"
  resolved "git://github.com/mishoo/UglifyJS2#278577f3cb75e72320564805ee91be63e5f9c806"
  dependencies:
    source-map "~0.5.1"
    yargs "~3.10.0"
  optionalDependencies:
    uglify-to-browserify "~1.0.0"
```
...
```
uglifyjs-webpack-plugin@^0.4.4, uglifyjs-webpack-plugin@^0.4.6:
  version "0.4.6"
  resolved "https://registry.yarnpkg.com/uglifyjs-webpack-plugin/-/uglifyjs-webpack-plugin-0.4.6.tgz#b951f4abb6bd617e66f63eb891498e391763e309"
  dependencies:
    source-map "^0.5.6"
    uglify-js "^2.8.22"
    webpack-sources "^1.0.1"
```

lili21:
That is the `yarn.lock` file. I won't say that is a workaround.

pkeuter:
[...]

PR #50, opened 2017-06-22
chore(package): explicit usage of `uglifyjs v2.0.0`

pierreneter:


sandfox:
I'm pretty sure that `^2.8.2` in semver already excludes `3.0.0.` and beyond.

https://jubianchi.github.io/semver-check/ has what I believe is an accurate tester.

pierreneter:
Opps. My fault, i think this will update to `uglify-js 3` when i run `npm update` but it not. Still `^2.8.29` is enough.
And one more, `uglify-js` required but did't install when i run `npm install uglifyjs-webpack-plugin --save` i think we need `uglify-js` in both `dependencies` and `peerDependencies`

pierreneter:
screen shot 2017-06-23 at 1 38 20 pm

package.json
```diff
@@ -42,11 +42,12 @@
   },
   "homepage": "https://github.com/webpack-contrib/uglifyjs-webpack-plugin",
   "peerDependencies": {
-    "webpack": "^1.9 || ^2 || ^2.1.0-beta || ^2.2.0-rc || ^3.0.0"
+    "webpack": "^1.9 || ^2 || ^2.1.0-beta || ^2.2.0-rc || ^3.0.0",
+		"uglify-js": "^2.8.29"
```
evilebottnawi: `uglify-js` already in dev deps, why your need peerDeps? This is unnecessary

pierreneter: let me show you why.

michael-ciniawsky:
@pierreneter Did you clean the cache and tried a 'fresh' install ? As others already stated the `semver` range is correct, but if there are still regressions we need to find a solution :)

pierreneter:
i think my problem is realated #52, or I have a big misunderstanding, but my project only work after i installed `uglify-js` at root of project. Maybe `uglify-js` in `dependencies` is redundant. I think my problem shall be done here. We can discuss about it in #52 or #49. Sorry for disturbing your time

michael-ciniawsky:
Alright closing in favour of #52 then :)

PR #53, opened 2017-06-22
refactor: performance optimizations

hulkish:


src/index.js
```diff
@@ -28,27 +60,38 @@ class UglifyJsPlugin {
 		compiler.plugin("compilation", (compilation) => {
 			if(options.sourceMap) {
 				compilation.plugin("build-module", (module) => {
-					// to get detailed location info about errors
+          // to get detailed location info about errors
```
Kovensky: bad indent :/

.eslintrc
```diff
@@ -6,6 +6,10 @@
 		"node": true,
 		"jasmine": true
 	},
+	"parser": "babel-eslint",
+	"parserOptions": {
+		"sourceType": "module",
```
Kovensky: all files here are scripts

src/index.js
```diff
@@ -28,27 +60,38 @@ class UglifyJsPlugin {
...
-				files.push.apply(files, compilation.additionalChunkAssets);
-				const filteredFiles = files.filter(ModuleFilenameHelpers.matchObject.bind(undefined, options));
-				filteredFiles.forEach((file) => {
-					const oldWarnFunction = uglify.AST_Node.warn_function;
-					const warnings = [];
-					let sourceMap;
+				const files = compilation.additionalChunkAssets;
+				let i = chunks.length - 1;
+				while(i--) {
+          // eslint-disable-next-line prefer-spread
```
Kovensky: bad indent here too

looks like the comments were indented with spaces while the rest of the file is indented with tabs?

src/index.js
```diff
@@ -97,136 +140,117 @@ class UglifyJsPlugin {
 								uglify.mangle_properties(ast, options.mangle.props);
 							}
 						}
-						const output = {};
-						output.comments = Object.prototype.hasOwnProperty.call(options, "comments") ? options.comments : /^\**!|@preserve|@license/;
+						const output = { ...options.output };
+						output.comments = options.comments || /^\**!|@preserve|@license/;
 						output.beautify = options.beautify;
```
Kovensky:
```js
const output = {
  comments: options.comments || /^\**!|@preserve|@license/,
  beautify: options.beautify,
  ...options.output
};
```

src/index.js
```diff
@@ -97,136 +140,117 @@ class UglifyJsPlugin {
...
-							// comments according to the two conditions
+              // Redefine the comments function to extract and preserve
+              // comments according to the two conditions
 							output.comments = (astNode, comment) => {
 								if(condition.extract(astNode, comment)) {
 									extractedComments.push(
-										comment.type === "comment2" ? "/*" + comment.value + "*/" : "//" + comment.value
-									);
+                    comment.type === "comment2" ? `/*${comment.value}*/` : `//${comment.value}`
+                  );
```
Kovensky: bad indent here too

src/index.js
```diff
@@ -28,27 +60,38 @@ class UglifyJsPlugin {
...
+          // eslint-disable-next-line prefer-spread
+					files.push.apply(files, chunks[i].files);
+				}
+
+				let file;
+				let sourceMap;
+				let input;
+				let inputSourceMap;
+				let map;
+				let oldWarnFunction;
```
Kovensky: Better to write a comment explaining why all these variables were hoisted outside the loop.

It's usually not a good idea to write variables outside the scope where they are needed. I assume this was done to avoid `transform-block-scoping` creating an anonymous function inside the loop for correct scoping.

src/index.js
```diff
@@ -60,17 +103,17 @@ class UglifyJsPlugin {
...
-									column: column
+									line,
+									column,
 								});
 								if(!original || !original.source || original.source === file) return;
 								if(!warningsFilter(original.source)) return;
-								warnings.push(warning.replace(/\[.+:([0-9]+),([0-9]+)\]/, "") +
-									"[" + requestShortener.shorten(original.source) + ":" + original.line + "," + original.column + "]");
+								warnings.push(`${warning.replace(WARN_PATTERN, "")
+                }[${requestShortener.shorten(original.source)}:${original.line},${original.column}]`);
```
Kovensky: also bad indent here

src/index.js
```diff
@@ -97,136 +140,117 @@ class UglifyJsPlugin {
...
+                    `/*! ${banner} */\n`, outputSource
+                  );
 								}
 							}
 						}
-						asset.__UglifyJsPlugin = compilation.assets[file] = outputSource;
-						if(warnings.length > 0) {
-							compilation.warnings.push(new Error(file + " from UglifyJs\n" + warnings.join("\n")));
+						compilation.assets[file] = outputSource;
+						asset.__UglifyJsPlugin = outputSource;
```
Kovensky: Looks like this should be a `WeakMap` instead.

But fixing this is outside the scope of this PR.

hulkish:
@Kovensky... was this switched to tabs recently? im pretty sure it's always been spaces during my working with this repo... only way I can think of them getting mixed like this would be when i pulled from master and resolved merge conflicts

Really hoping tabs are not intentionally preferred...

d3viant0ne:
Everything in contrib is spaces

Kovensky:
Looks like the .eslintrc in this repo needs to be fixed, then. It's specifying tab indent.

d3viant0ne:
@Kovensky - Should be taken care of with the defaults PR.

d3viant0ne:
@hulkish - It's usually helpful to put at least a brief description in the original comment outlining the reason for the change / what problem you are solving & the approach you took.

src/index.js
```diff
@@ -12,6 +13,33 @@ const RequestShortener = require("webpack/lib/RequestShortener");
...
+			if(condition[key] === "all") {
+				condition[key] = () => true;
+				break;
+			}
+			const regex = new RegExp(condition[key]);
+			condition[key] = (astNode, comment) => regex.test(comment.value);
+			break;
+		}
+		default: {
+			const defaultRegex = condition[key];
```
```
defaultRegex => regex 
```

Lexical 'Block' Scope ?  Not :100: && Nitpick.. 😛

src/index.js
```diff
@@ -22,33 +50,53 @@ class UglifyJsPlugin {
...
-					const oldWarnFunction = uglify.AST_Node.warn_function;
-					const warnings = [];
-					let sourceMap;
+				const warnings = [];
+				const output = {
+					comments: options.comments || /^\**!|@preserve|@license/,
+					beautify: options.beautify,
+					...options.output
+				};
+				const outputSourceSet = new WeakSet();
```
michael-ciniawsky: `WeakSet()` as sad as it would be a `semver` major 😉

src/index.js
```diff
@@ -121,27 +165,8 @@ class UglifyJsPlugin {
...
-										}
-										var regex = new RegExp(condition[key]);
-										condition[key] = (astNode, comment) => regex.test(comment.value);
-										break;
-									default:
-										regex = condition[key];
-										condition[key] = (astNode, comment) => regex.test(comment.value);
-								}
-							});
+							ensureConditionFunc(condition, "preserve");
```
michael-ciniawsky: Please `-...-->Func<--` the function call should speak for itself :D, just `ensure(condition, ...)` maybe ?

src/index.js
Kovensky: That's how you write a switch case when you need a lexical declaration.

This has precedent in C/C++ as well.

hulkish:
After creating the test PR and running those tests on this PR... I realized i had bugs which was causing the disproportionate performance gain i saw originally.

Therefore... I'm going to abandon this because after fixing said bugs there was still a performance gain - but too tiny to really be valueable.

🔥 this pr!

PR #54, opened 2017-06-22
refactor: upgrade to `uglifyes v3.0.0`

graingert:


michael-ciniawsky:
There is `uglifyes-webpack-plugin` and aren't there API breaking changes betwen `uglifyes` (`v3.0.0`) && `uglifyjs` (`v2.0.0`) which makes this more complicated then switching the dependency?

graingert:
I was just giving it a go to see what tests pass, but yeah there's none

michael-ciniawsky:
Superseded by #55 Thx :)

Issue #1, opened 2017-01-21
Port uglifyjs tests from the main repository


bebraw:
The current code is missing tests. Those should be ported over from the main project.

michael-ciniawsky:
Closed by #56

PR #56, opened 2017-06-24
test: add && backport tests from webpack core

hulkish:
Closes https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/1. This is a port of the [existing UglifyJsPlugin tests](https://github.com/webpack/webpack/blob/master/test/UglifyJsPlugin.test.js). 

I've converted them to jest format and also added a snapshot test. I only added one snapshot test for now.


package.json
```diff
@@ -28,11 +28,15 @@
   ],
   "jest": {
     "collectCoverage": true,
-    "moduleFileExtensions": [
-      "js"
-    ],
-    "moduleDirectories": [
-      "node_modules"
+	  "coverageDirectory": "./coverage/",
```
michael-ciniawsky: Are most of these not `defaults` anyways? Also the question if we should add this to `webpack-defaults` then

__tests__/helpers.js
```diff
@@ -0,0 +1,47 @@
...
+		};
+	}
+
+	getEventBindings() {
+		return this.events;
+	}
+};
+
+
+exports.runCompiler = function runCompiler(compiler) {
```
michael-ciniawsky: Why not `function compile (fixture, config) {}` [Discuss], so

```js
import { compile } from './helpers/webpack.js'
import UgilfyJS from '..'

test('${name}', () => {
   const config = { plugins: [ new UgilfyJS() ] }

   compile(`${fixture}`, config).then((result) => {
      // ...
   }) 
})
```

__tests__/UglifyJsPlugin.test.js
```diff
@@ -0,0 +1,824 @@
+"use strict";
```
michael-ciniawsky: Possible to split these into separate files ? e.g

```
|– test
| |– __snapshots__
| |– index.test.js  (Basic, with defaults)
| |– options1.test.js
| |– options2.test.js
| |– .... 
```

package.json
```diff
@@ -53,20 +57,21 @@
...
     "babel-eslint": "^7.1.1",
-    "babel-jest": "^18.0.0",
+    "babel-jest": "^20.0.3",
     "babel-plugin-syntax-object-rest-spread": "^6.13.0",
     "babel-plugin-transform-object-rest-spread": "^6.20.2",
     "babel-preset-es2015": "^6.18.0",
     "eslint": "^3.13.1",
     "eslint-config-airbnb": "^14.0.0",
     "eslint-plugin-import": "^2.2.0",
+    "eslint-plugin-jest": "^20.0.3",
```
michael-ciniawsky: Needed ? Also needs to be in snyc with `webpack-defaults` then :)

package.json
ev1stensberg: `coverageDirectory` defaults to `coverage`

package.json
ev1stensberg: I don't fully know if this is needed, as most of these are done during dev and caught up during PR reviews/submits:
skjermbilde 2017-06-25 kl 19 52 01

__tests__/helpers.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+
+const MemoryFileSystem = require("memory-fs");
+const webpack = require("webpack");
+
+exports.PluginEnvironment = class PluginEnvironment {
```
hulkish: from [here](https://github.com/webpack/webpack/blob/master/test/helpers/PluginEnvironment.js)

The `memory-fs` is being used to run a webpack compile without needing to emit files for the test.

__tests__/helpers.js
hulkish: Well, I'm ok with either direction. I hadn't added many snapshots tests yet - but I built this with the idea of making many in mind and thought it would be useful.

package.json
hulkish: I added `coverageDirectory` to try and get the codecov ci check to succeed. Right now my best guess is that it's only not succeeding because it compares the codecov to master branch - which currently has no tests. Please advise.

package.json
hulkish: One issue I can into is that `jest` needs to be a global. I can remove this plugin and just allow jest to be a global, I suppose.

codecov.yml
```diff
@@ -0,0 +1,17 @@
+codecov:
```
hulkish: wouldn't `webpack-defaults` just create this file here anyway? Or, do u mean it also needs to be added there?

__tests__/UglifyJsPlugin.test.js
hulkish: Sure, but can you clarify? Do u mean put snapshot tests in diff file? Or have the snapshots themselves be separated?

__tests__/helpers.js
michael-ciniawsky: This is enterily open for discussion and maybe out of scope of this PR, the direction is fine and changing it in a following PR is totally ok

codecov.yml
michael-ciniawsky: yep, if we need `codecov.yml` for coverage to work correctly, we should add it to `webpack-defaults` aswell to keep repos in sync

__tests__/UglifyJsPlugin.test.js
michael-ciniawsky: The `__snapshots__` directory is just in there as a 'where are we' orientation point, `jest` will create a separate `snapshot` for each `${name}.test.js` in `__snapshots__`. The important part is to split test into logical groups, so particular parts can be tested more independently

__tests__/helpers.js
michael-ciniawsky: 👍 Can we make this more generic and could you elaborate a bit more in detail about it, I would like to write docs about it then, for the `webpack-defaults` test boilerplate 

`ExpectedPlugin` `TestPlugin` `JestPlugin`?

package.json
michael-ciniawsky: kk I can't really say, bc I don't use it myself and as a StandardJS user I never really dealt with ESLint aswell :), if ESLint complains it needs t be solved either in `eslint-config-webpack` or via a PR adding this plugin to `webpack-defaults`in te long-term

__tests__/helpers.js
michael-ciniawsky: Would it also be possible to get the `MemoryFS` output 'bundle' and snapshot test it? 

```js
test('${name}', () => {
   compiler(fixture, config)
     .then((result) => {
        result.stats
        result.output // MemoryFS Instance as `bundle.js`
      })
     .catch((err) => err)
})

Comment (hulkish) [inline __tests__/helpers.js]: Yes, and that's probably better anyway. Im thinking that's probably something for a new PR, though - since it will involve probably slightly more tooling to be created for that.

Comment (michael-ciniawsky) [inline __tests__/helpers.js]: How does `result.output` look like in (terms of noise), could you post it here when reddy ? 😛 
If we agree on 'extending' a base config we could evetually add `CommonChunksPlugin`to the base to split the webpack runtime by default

```js
[ 
  new CommonChunksPlugin({ 
    name: [ 'runtime', `${fixture}` ], 
    minChunks: Infinity 
  }) 
].concat(config.plugins || [])
```

```js
webpackJsonp([`${fixture}`],{...bundle}
```

hulkish:
@michael-ciniawsky better?

__tests__/UglifyJsPlugin.test.js
ev1stensberg: You could look at webpack-cli to see how we layered our tests :)

__tests__/empty-options.snapshot.test.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+const UglifyJsPlugin = require("../src/index");
+const createCompiler = require("./helpers").createCompiler;
+const compile = require("./helpers").compile;
+
+describe("snapshotss", () => {
```
michael-ciniawsky: `snapshotss` => `snapshots`

hulkish:
added https://github.com/webpack-contrib/webpack-defaults/pull/77 for `codecov.yml`

michael-ciniawsky:
`travis.yml` needs an update for `node `

```diff
node_js:
- - "4"
- - "5"
- - "6"
+ -node
+ - 6
+ - 4
```

`invalid-options.test.js` is failing due to path mismatches in the Error StackTrace, the individual part could normally be replaced with a regex `.*\/...` but this one is hairy to substitute 😛

__tests__/invalid-options.test.js
```diff
@@ -15,8 +16,11 @@ describe("when applied with invalid options", () => {
 		}).apply(compiler);
 
 		return compile(compiler).then((stats) => {
-			expect(stats.compilation.warnings).toMatchSnapshot("warnings");
-			expect(stats.compilation.errors).toMatchSnapshot("errors");
+			const errors = stats.compilation.errors.map(cleanErrorStack);
```
michael-ciniawsky: clever :)

hulkish:
@michael-ciniawsky interesting... it fails in ci on node 4 only: https://travis-ci.org/webpack-contrib/uglifyjs-webpack-plugin/jobs/247753398#L242

michael-ciniawsky:
Can you remove the `yarn.lock` file please `webpack-defaults` uses `package-lock.json`&& `npm@5` :)

PR #55, opened 2017-06-24
feat: update to `uglify-es`

Hartorn:
I tried this, because I wanted to be able to use webpack, with uglify es and not js.

Basically, I moved the options for Uglify into uglifyOptions key, and added an option for ES5 or ES6.

I also changed the code, to use the uglify.minify method.

Not sure if I understood something wrong, but I hope this can help. 
I can correct this if needed (tested on some of my projects, and it seemed to work)

EDIT : 
Here is a quick sum up of the modifications, please ask for precisions if needed.

 * Uglify now expose a single high level function `minify(code, options)`, which go through all the steps of minification (parse, compress, mangle, output)
 * it does not throw errors anymore, so the try-catch is not needed (test if there is an error key in the return of the function)
 * it is possible to get the warnings in the result also, so it is not needed anymore to mingle with AST_NODE `warn_function`
 * it is possible to give a filter for comments, so we can hook on that for exporting comments
 * it is possible to give a preamble in the output, so we can use that for the banner thing (License file, comments, ...)

Basically, we do not need to use any other function than minify from uglify, it is mainly about giving proper options, handle the source maps and output.

I did not had time to update the docs, will do in the next days.

PR #57, opened 2017-06-29
refactor: changed `__UglifyJsPlugin` dereference to use `WeakSet` instead

hulkish:
So, as a proof of concept... I created this: https://github.com/hulkish/dummy

Pull that repo, then run:

```
yarn
node scripts/profile.js
```

You will see an output which looks like:
```
untouched_node@v6.11.0 compile...
untouched_node@v6.11.0 files length 68
untouched_node@v6.11.0 profiler.startProfiling()
untouched_node@v6.11.0: 27549.784ms
untouched_node@v6.11.0 profiler.stopProfiling()
untouched_node@v6.11.0 profile.export()
untouched_node@v6.11.0 profile saved: /Users/shargrove/dev/repos/dummy/untouched_node@v6.11.0.uglifyjs-webpack-plugin.cpuprofile
forEach_WeakSet_node@v6.11.0 compile...
forEach_WeakSet_node@v6.11.0 files length 68
forEach_WeakSet_node@v6.11.0 profiler.startProfiling()
forEach_WeakSet_node@v6.11.0: 14367.834ms
forEach_WeakSet_node@v6.11.0 profiler.stopProfiling()
forEach_WeakSet_node@v6.11.0 profile.export()
forEach_WeakSet_node@v6.11.0 profile saved: /Users/shargrove/dev/repos/dummy/forEach_WeakSet_node@v6.11.0.uglifyjs-webpack-plugin.cpuprofile
```

then, you can compare the output of these 2 folders:
- `dist/untouched_node@v6.11.0`
- `dist/forEach_WeakSet_node@v6.11.0`

The output is the same, but look at the significant difference in time.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,14 +50,14 @@ class UglifyJsPlugin {
         // eslint-disable-next-line prefer-spread
         files.push.apply(files, compilation.additionalChunkAssets);
         const filteredFiles = files.filter(ModuleFilenameHelpers.matchObject.bind(undefined, options));
+        const uglifiedAssets = new WeakSet();
         filteredFiles.forEach((file) => {
           const oldWarnFunction = uglify.AST_Node.warn_function;
           const warnings = [];
           let sourceMap;
           try {
             const asset = compilation.assets[file];
-            if (asset.__UglifyJsPlugin) {
-              compilation.assets[file] = asset.__UglifyJsPlugin;
+            if (uglifiedAssets.has(asset)) {
               return;
             }
             let input;
@@ -220,9 +220,7 @@ class UglifyJsPlugin {
                 }
               }
             }
-            compilation.assets[file] = outputSource;
-            asset.__UglifyJsPlugin = outputSource;
-
+            uglifiedAssets.add(compilation.assets[file] = outputSource);
             if (warnings.length > 0) {
               compilation.warnings.push(new Error(`${file} from UglifyJs\n${warnings.join('\n')}`));
             }


michael-ciniawsky:
Change the base branch back to `master` please, since it seem to landed there yet :)

Kovensky:
Interesting, I didn't expect using a `WeakSet` to actually improve performance by much; it just "felt right" considering how the extra key was added and used. But I guess the cost of forcing a hidden class change by adding an extra key was high enough.

mikesherov:
@Kovensky it's not that WeakSet beats prop assignment that makes this faster! It's:

img_4037
img_4038
img_4039
img_4040
img_4041
img_4042

mikesherov:
@Kovensky that is to say, this was the old code:

```
asset = compilation.assets[file]
if(asset.__UglifyJsPlugin) {
	compilation.assets[file] = asset.__UglifyJsPlugin;
	return;
}
outputSource = uglify();
asset.__UglifyJsPlugin = compilation.assets[file] = outputSource;
```

So there's a circular set of references there.

hulkish:
@mikesherov Another thing to consider here... what if we just filtered the files array to only contain unique values? Wouldn't that have same effect?

mikesherov:
Oh interesting, the weakSet is in the wrong spot? Can different compilations have the same asset?

hulkish:
@mikesherov what u mean wrong spot?

mikesherov:
Can be placed at RequestShortener's scope?

hulkish:
You're right... I considered that but i wonder if in principle it's actually better to keep it in 'optimize-chunk-assets` handler scope... For ex... i thought about how this might behave in watch mode and wondered if this would cause an issue.

mikesherov:
It can't *possibly*, it's a WeakSet! It won't leak memory, and a match will only happen when it truly should happen!

hulkish:
Ah, very good point.

michael-ciniawsky:
Sry for the heads up, but could we triage landing/ not landing `uglify-es` (#55) first before optimizing further 🙃 ?

mikesherov:
YESSSSS

michael-ciniawsky:
ALRIGHTTTT 😛 :D

Issue #32, opened 2017-05-10
Support UglifyJS 3


bebraw:
Currently using UglifyJS 3 breaks the plugin. It has a new API so the plugin has to change accordingly.

rainjay:
if u are using yarn, paste the codes below in yarn.lock
```
"uglify-js@git://github.com/mishoo/UglifyJS2#harmony":
  version "3.0.0"
  resolved "git://github.com/mishoo/UglifyJS2#3fac29a01787b07ce9e43fb87a17bde33123509d"
  dependencies:
    commander "~2.9.0"
    source-map "~0.5.1"
  optionalDependencies:
    uglify-to-browserify "~1.0.0"
```
Replace the old one, yarn and then it will work, Although it is not recommended to edit in yarn.lock.
Just a temporary solution.

t47io:
I see this error:
```
Error: Cannot find module 'uglify-js'
```
Need to rename to `'uglify-es'`?

bebraw:
`uglify-es` won't work with the current version of the plugin without serious changes.

rainjay:
another way: paste ` "uglify-js": "git://github.com/mishoo/UglifyJS2#harmony-v2.8.22" ` in your devDependency in package.json

ro-savage:
@bebraw - Is this something that is currently being worked on?

I had a look at the source to see if I could do a PR to support UglifyJS3 but it looks like this is using a lot of undocumented and internal APIs of UglifyJS2 and felt like it would be almost a total rewrite.

Maybe @alexlamsl or @kcz could help?

bebraw:
@ro-savage I don't think anyone is working on this at the moment as people are busy with other things. It's a big rewrite, yeah.

noBlubb:
I created an updated fork two days ago, that has been forked and improved again here https://github.com/Nckcol/uglify-es-webpack-plugin and is now almost feature complete.

bebraw:
@noBlubb Do you want to look into merging the work back here or do you prefer to maintain it on your own? Webpack main repository has some Uglify tests that have to be ported over and adjusted.

noBlubb:
I would like to transfer the plugin back to webpack-contrib, as I do not feel ready to maintain this all by myself. This started of as some kind of POC for me and I'm happy to give something back to the webpack community :)

bebraw:
[...]

PR #61, opened 2017-06-29
docs(README): standardize

michael-ciniawsky:
[README](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/readme)

- [x] Styling && Format
- [x] `uglify-es` changes (`options.uglifyOptions`)

Closes #27

Issue #33, opened 2017-05-10
switch to uglify-es


graingert:
Related somewhat to https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/32

get rid of all the "if using ES2017 install from git" etc, and switch to `uglify-es` only.

skorunka:
How can I use `uglify-es` instead of `uglify-js` in webpack?

bebraw:
@skorunka It's not possible until someone ports the plugin to the new API of Uglify.

skorunka:
@bebraw I see, is there any workaround?

bebraw:
@skorunka No workaround until someone has time to perform the upgrade,

skorunka:
@bebraw Thank you.

johnspurlock:
@skorunka Hmm, I've been using a similar workaround (see below) while waiting for this issue, and it seems to be working for me.  Copying here in case it helps you.

```javascript
/* package.json */
"devDependencies": {
    "uglify-js": "git://github.com/mishoo/UglifyJS2#harmony-v2.8.22",
    "uglifyjs-webpack-plugin": "0.4.3",
```
```javascript
/* webpack.config.js */
const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
new UglifyJSPlugin()
```

skorunka:
That helped! I had to delete `node_modules` folder and `npm i` so I get the rid of all previously installed `uglify-js` mutations.

JemarJones:
Why does the README tell us to use github.com/mishoo/UglifyJS2#harmony and not mention @johnspurlock's fix?

bebraw:
@JemarJones PR welcome. Thanks. 👍

zhaiduo:
Try method from @johnspurlock, still not work.

> ERROR in scripts/p.js from UglifyJs
Unexpected token: punc ()) [scripts/p.js:80,64]

This is my devDependencies:

> "babel-cli": "^6.24.1",
    "babel-core": "^6.25.0",
    "babel-preset-es2015": "^6.24.1",
    "uglify-js": "git://github.com/mishoo/UglifyJS2#harmony-v2.8.22",
    "uglifyjs-webpack-plugin": "^0.4.3",
    "webpack": "^2.6.1",
    "webpack-bundle-analyzer": "^2.8.2",
    "webpack-dev-middleware": "^1.10.2",
    "webpack-hot-middleware": "^2.18.0",

[...]

PR #63, opened 2017-06-30
feat: update to `uglify-es`

Hartorn:
@michael-ciniawsky @hulkish @graingert : Here is the new PR.

This PR is about moving to uglify-es (following of #55, due to messy chained rebase/push-force, etc...)

Here is a quick sum up of the modifications, please ask for precisions if needed.

 * Uglify now expose a single high level function `minify(code, options)`, which go through all the steps of minification (parse, compress, mangle, output)
 * it does not throw errors anymore, ~~so the try-catch is not needed (test if there is an error key in the return of the function)~~
 * it is possible to get the warnings in the result also, so it is not needed anymore to mingle with AST_NODE `warn_function`
 * it is possible to give a filter for comments, so we can hook on that for exporting comments
 * it is possible to give a preamble in the output, ~~so we can use that for the banner thing (License file, comments, ...)~~

Basically, we do not need to use any other function than minify from uglify, it is mainly about giving proper options, handle the source maps and output.

I did not had time to update the docs, I will do in the next days.

TODO : 

 [ ] Update the documentation
 [ ] Add new test for the new functionnalities


Hartorn:
@hulkish Honestly, I'm not so familiar with unit testing like this. I corrected the existing ones (only the configuration given to the plugin). I might need help on this, if anyone wants to lend a hand :)

michael-ciniawsky:
I take care of the docs in #61 after this PR :)

cc @hulkish for Unit test

hulkish:
@Hartorn sure, you wanna give me access to your fork and ill add the test.

hulkish:
@Hartorn Nvm, I see you've already included your updates to the tests: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/63/files

Hartorn:
@hulkish @michael-ciniawsky Already invited you both to my fork, maybe it will be useful.

I pushed the modifications to the package-lock, did not pay attention the repo was in npm 5.
Build should be all good

@hulkish I thought you were talking about adding new cases for uglify-es options. 
I might still need to add a check on source map (the one inside uglifyOptions must be ignored, we only care about the one in options). 
Thanks for checking :) 👍

src/index.js
```diff
@@ -3,247 +3,268 @@
...
+    this.uglifyOptions = this.options.uglifyOptions || {};
+  }
+
+  static buildDefaultUglifyOptions({ ecma, warnings, parse, compress, mangle, output, toplevel, ie8 }) {
+    return {
+      ecma,
+      warnings,
+      parse: parse || {},
+      compress: compress || {},
+      /* eslint-disable no-undefined */
```
graingert: eslint-disable-next-line

test/all-options.test.js
```diff
@@ -33,6 +27,14 @@ describe('when applied with all options', () => {
           return `License information can be found in ${licenseFile}`;
         },
       },
+      uglifyOptions: {
```
michael-ciniawsky: Possible to keep the individual options in the top level e.g via `filter()` ?

src/index.js
```diff
@@ -3,247 +3,267 @@
...
+  }
+
+  static buildDefaultUglifyOptions({ ecma, warnings, parse, compress, mangle, output, toplevel, ie8 }) {
+    return {
+      ecma,
+      warnings,
+      parse: parse || {},
+      compress: compress || {},
+      // eslint-disable-next-line no-undefined
+      mangle: mangle === undefined || mangle === null ? true : mangle,
```
graingert: mangle == null is idiomatic

test/all-options.test.js
Hartorn: I felt like they were too many options to keep them at top level. (https://github.com/mishoo/UglifyJS2/tree/harmony#minify-options-structure)
That way, I thought it was easier to understand what's coming from the plugin, or what's coming from UglifyJs, and so easier to read the docs according to the need.

michael-ciniawsky: hmm.. I basically agree 😛, but `uglifyOptions` is too verbose. What's your opinion on remvoing `warningFilters` && `extractComments` ? 🙃

Hartorn: `warningFilters` : I think it can be used by some people to filter for example on node_modules and still see warnings about their code, maybe to check dead code presence, or optimisation.

`extractComment`: It is for copyright/license reason I think => that way, the file you load is steel small, plus a preamble redirecting to another file containing all the copyrights/licenses.

michael-ciniawsky: Alright

michael-ciniawsky:
What are the `uglifyOptions` defaults atm => `new UglifyJSPlugin()` ?

Hartorn:
options.test = /\.js($|\?)/i;
options.warningsFilter =  (() => true);
options.comments = /^\**!|@preserve|@license/;

Since the test on compress and mangle are `!== false`, they are also activated by default.
That's it I think

(Have to go now, we'll come back tomorrow)

michael-ciniawsky:
kk thx

hulkish:
@Hartorn @michael-ciniawsky how's that?

michael-ciniawsky:
👍

test/es2015-module-options.test.js
```diff
@@ -0,0 +1,155 @@
...
+  compile,
+} from './helpers';
+
+describe('when applied with uglifyOptions.ecma', () => {
+  it('matches snapshot for ecma 5', () => {
+    const compiler = createCompiler({
+      entry: `${__dirname}/stubs/es-entry.js`,
+      output: {
+        path: `${__dirname}/dist-es`,
+        filename: '[name].[chunkhash].js',
```
michael-ciniawsky: Are hashes needed ?

test/es2015-module-options.test.js
```diff
@@ -0,0 +1,155 @@
+import UglifyJsPlugin from '../src/index';
```
michael-ciniawsky: `es2015-module-options` => `es2015`

test/helpers.js
```diff
@@ -31,20 +31,30 @@ exports.compile = function compile(compiler) {
...
 
-exports.createCompiler = function createCompiler(options) {
-  const compiler = webpack(options || {
-    bail: true,
+exports.createCompiler = function createCompiler(options = {}) {
+  const compiler = webpack({
     cache: false,
     entry: `${__dirname}/stubs/entry.js`,
+    resolve: {
+      unsafeCache: false,
```
michael-ciniawsky: What does this exactly and is it needed ? 🙃

test/stubs/es-dep.js
```diff
@@ -0,0 +1,2 @@
+export const bar = 'bar';
```
michael-ciniawsky: `stubs` => `fixtures`

michael-ciniawsky: `fixtures/es-dep.js` => `fixtures/es2015/dep.js`

test/stubs/es-entry.js
```diff
@@ -0,0 +1,15 @@
+import foo, { bar } from './es-dep';
```
michael-ciniawsky: `fixtures/es-entry.js` => `fixtures/es2015/entry.js`

test/helpers.js
hulkish: Probably not needed... I added it to debug an issue while i was making the es test.. Ill go ahead and remove it.

michael-ciniawsky: If it is helpful for debugging and less likely to introduce conflicts it's ok, but if it doesn't affect unit tests better reduce the noise where possible 🐤

test/es2015-module-options.test.js
michael-ciniawsky: Can this possibly conflict in future CI runs without any direct changes ?

hulkish: [chunkhash] help with unique file names for the snapshotted files tests l, reads better when they change

hulkish: dont think this will cause ci issues, since everything is passing it seems.

Hartorn:
@hulkish Looks good to me :) 👍

src/index.js
```diff
@@ -3,247 +3,267 @@
...
-          const oldWarnFunction = uglify.AST_Node.warn_function;
-          const warnings = [];
-          let sourceMap;
-          try {
+        chunks.reduce((acc, chunk) => acc.concat(chunk.files || []), [])
+          .concat(compilation.additionalChunkAssets || [])
+          .filter(ModuleFilenameHelpers.matchObject.bind(null, this.options))
+          .forEach((file) => {
+            // Copy uglify options
+            const uglifyOptions = UglifyJsPlugin.buildDefaultUglifyOptions(this.uglifyOptions);
```
mikesherov: do you have to build this each iteration?

src/index.js
```diff
@@ -3,247 +3,267 @@
...
-        // eslint-disable-next-line prefer-spread
-        files.push.apply(files, compilation.additionalChunkAssets);
-        const filteredFiles = files.filter(ModuleFilenameHelpers.matchObject.bind(undefined, options));
         const uglifiedAssets = new WeakSet();
-        filteredFiles.forEach((file) => {
-          const oldWarnFunction = uglify.AST_Node.warn_function;
-          const warnings = [];
-          let sourceMap;
-          try {
+        chunks.reduce((acc, chunk) => acc.concat(chunk.files || []), [])
```
mikesherov: when would chunk.files ever not be an array?

src/index.js
Hartorn: @graingert I don't understand this. What you mean is that I should not test `mangle === null` but only ` mangle === undefined ? true : mangle` ?

src/index.js
```diff
@@ -3,247 +3,267 @@
   Author Tobias Koppers @sokra
 */
 
-// TODO: temporarily disabled rules
-/* eslint-disable
-  no-undefined,
-  no-param-reassign,
```
mikesherov: We shouldn't get rid of this. `no-param-reassign` is a bad rule, IMO. It makes us have to do confusing things like

```
+        compilation.plugin('build-module', (moduleArg) => {
            // to get detailed location info about errors
 -          module.useSourceMap = true;
 +          const moduleVar = moduleArg;
 +          moduleVar.useSourceMap = true;
          });
```

src/index.js
```diff
@@ -3,247 +3,267 @@
...
+    this.options.warningsFilter = this.options.warningsFilter || (() => true);
+
+    this.uglifyOptions = this.options.uglifyOptions || {};
+  }
+
+  static buildDefaultUglifyOptions({ ecma, warnings, parse, compress, mangle, output, toplevel, ie8 }) {
+    return {
+      ecma,
+      warnings,
+      parse: parse || {},
```
mikesherov: switch to a default parameter

src/index.js
```diff
@@ -3,247 +3,267 @@
...
+      compress: compress || {},
```
mikesherov: switch to a default parameter

src/index.js
```diff
@@ -3,247 +3,267 @@
...
+    return new Error(`${file} from UglifyJs\n${err.stack}`);
+  }
 
-    this.options = options;
+  static buildWarnings(warnings, file, sourceMap, warningsFilter, requestShortener) {
+    return warnings.reduce((accWarnings, warning) => {
+      if (!sourceMap) {
+        accWarnings.push(warning);
+      } else {
+        const match = /\[.+:([0-9]+),([0-9]+)\]/.exec(warning);
```
mikesherov: can you hoist this regex? Used in two spots in a loop.

src/index.js
```diff
@@ -3,247 +3,267 @@
...
+    } else if (err.msg) {
+      return new Error(`${file} from UglifyJs\n${err.msg}`);
     }
+    return new Error(`${file} from UglifyJs\n${err.stack}`);
+  }
 
-    this.options = options;
+  static buildWarnings(warnings, file, sourceMap, warningsFilter, requestShortener) {
+    return warnings.reduce((accWarnings, warning) => {
+      if (!sourceMap) {
```
mikesherov: this should likely move out of the loop...

```
if (!sourceMap) {
 return warnings;
}

return warnings.reduce...
```

src/index.js
graingert: `mangle === undefined || mangle === null` is the same as `mangle == null`

src/index.js
```diff
@@ -3,247 +3,267 @@
...
             }
-            let input;
-            let inputSourceMap;
-            if (options.sourceMap) {
-              if (asset.sourceAndMap) {
-                const sourceAndMap = asset.sourceAndMap();
-                inputSourceMap = sourceAndMap.map;
-                input = sourceAndMap.source;
+
+            try {
```
mikesherov: is there a reason to `try` anymore? I thought Uglify 3 doesn't throw? Is that not true?

src/index.js
```diff
@@ -3,247 +3,267 @@
...
-              } else if (Object.prototype.hasOwnProperty.call(options.extractComments, 'condition')) {
-                // Extract condition is given in extractComments.condition
-                condition.preserve = output.comments;
-                condition.extract = options.extractComments.condition;
+
+              // Calling uglify
+              const { error, map, code, warnings } = uglify.minify({ [file]: input }, uglifyOptions);
+
+              // Handling results
+              if (error) {
```
mikesherov: instead of `throw` and immediate `catch`, perhaps you can do:

```
if (error) {
  compilation.errors.push(UglifyJsPlugin.buildError(error, file, sourceMap, compilation, requestShortener));
  return;
}
```

here. This will avoid a try/catch deoptimization until everyone is on Node 8 with V8 that has TF+I that can properly optimize try/catch statements.

src/index.js
Hartorn: It is true that uglify does not throw. 
I had to put back the try catch to handle wrong asset, for example if an asset.source is not a function.
Tests were failing because of that.

src/index.js
Hartorn: Thks for the explanation.

src/index.js
Hartorn: Yeah, I thought it was weird. I will put it back.

src/index.js
Hartorn: I guess that's a bit overkill, it's an habit (bad ?) I took, since I saw people not use empty array but null or something like. 
Would this kind of defaulting have negative effects on performance or side effects ?

src/index.js
Hartorn: I guess not. I am reassigning only a few properties. Will change it.

src/index.js
mikesherov: is there any other way to avoid the try catch?

src/index.js
mikesherov: no, but if it's unnecessary it's unnecessary. It adds intellectual overhead when I go to read the code... "oh shit, when is this the case?"

src/index.js
mikesherov: if `asset.source` is not a function, can't we add the error way earlier? Getting rid of the try/catch is going to be a big win.

src/index.js
Hartorn: Ok then, will remove it :)

src/index.js
Hartorn:
``` js 
              if (this.options.sourceMap) {
                if (asset.sourceAndMap) {
                  const sourceAndMap = asset.sourceAndMap();
                  inputSourceMap = sourceAndMap.map;
                  input = sourceAndMap.source;
                } else {
                  inputSourceMap = asset.map();
                  input = asset.source();
                }
                sourceMap = new SourceMapConsumer(inputSourceMap);
                // Add source map data
                uglifyOptions.sourceMap = {
                  content: inputSourceMap,
                };
              } else {
                input = asset.source();
              }
```
The problem is this snippet. I can isolate it in a function, and only try catch this block, but I dont see a solution to drop the try catch.

Even if I try to check if source, map, sourceAndMap exist and are functions, that does not guarantee a no exception call.

Hartorn: Just thought of it : actually, every call to user given function (warningsFilter, commentsFile, banner) should be try-catched.

Hartorn: @mikesherov I made this, if you want to have a look, but I'm not so convinced. https://gist.github.com/Hartorn/14cba0e11382c38a91c6ac1b0c0cc9c1

The question is :  if the user give wrong banner, warningsFilter or extractComments functions, should it crash, or go on the compilation ?

I'm not sure it's going to be a big win, since Uglify is using inside a try-catch : https://github.com/mishoo/UglifyJS2/blob/master/lib/minify.js#L49

mikesherov: So I don't think we should be try-catching user supplied functions. That's not a common paradigm. User is responsible for bringing a plugin that doesn't throw, but that's separate discussion.

You're right about uglify try/catching though. Leave it as is :-/

src/index.js
```diff
@@ -3,247 +3,277 @@
...
-    if (typeof options.compressor !== 'undefined') {
-      options.compress = options.compressor;
+
+    this.options.test = this.options.test || /\.js($|\?)/i;
+    this.options.warningsFilter = this.options.warningsFilter || (() => true);
+
+    this.uglifyOptions = this.options.uglifyOptions || {};
+  }
+
+  static buildDefaultUglifyOptions({ ecma, warnings, parse = {}, compress = {}, mangle, output, toplevel, ie8 }) {
```
michael-ciniawsky: `buildDefaultUglifyOptions` => `defaults`

src/index.js
```diff
@@ -3,247 +3,277 @@
...
+    } else {
+      this.options = options || {};
     }
-    if (typeof options.compressor !== 'undefined') {
-      options.compress = options.compressor;
+
+    this.options.test = this.options.test || /\.js($|\?)/i;
+    this.options.warningsFilter = this.options.warningsFilter || (() => true);
+
+    this.uglifyOptions = this.options.uglifyOptions || {};
```
michael-ciniawsky: `this.options.uglifyOptions` => `this.options.uglify` (?)

src/index.js
```diff
@@ -3,247 +3,277 @@
...
+      mangle: mangle == null ? true : mangle,
+      // Ignoring sourcemap from options
+      sourceMap: null,
+      output: { ...defaultUglifyOptions.output, ...output },
+      toplevel,
+      ie8,
+    };
+  }
+
+  static buildError(err, file, sourceMap, requestShortener) {
```
michael-ciniawsky: `this.buildError`  => `this.error`

src/index.js
hulkish: i actually like @Hartorn approach here because its more clear of an indicator *hey... these are options sent straight to uglify"

michael-ciniawsky: yeah.. it is suboptimal in both ways tbh 🙃

src/index.js
michael-ciniawsky: `this.buildWarnings` => `this.warn`

src/index.js
hulkish: maybe a better way is to introduce an `UglifyjsPreset` config object that we can build-in for most common uses cases?  for ex:

```
plugins: [
  // option #1
  UglifyjsPlugin.fromPreset('default-web')

  // option #2
  new UglifyjsPlugin({
    preset: 'default-web`,
    sourceMap: true
  })
]
```

michael-ciniawsky: No presets please, that would be another webpack specific thing to learn then, I'm meticulous here already, but better triage the simplest way, when we can 😛

d3viant0ne: 1.) The plugin should consume configurations, not offer them. Too many edge cases and more maintenance then you would think. The behavior of the plugin configuration should remain consistent with the existing major version range in this regard.

2.) Use `uglifyOptions `.

d3viant0ne:
Has anyone tried perf testing this, even in a limited fashion?

I understand it's Apples to Volkswagens when trying to compare 2.x with ES but it wouldn't hurt to run the same project through both to make sure performance doesn't go down hill.

Nothing really stands out that screams `perf sucking black hole` but a sanity check with a prod(ish) application certainly can't hurt.

odensc:
Hey @d3viant0ne,

I've been using this PR in my build config for a while, and it works - but after seeing your comment I also became interested in the performance hit/benefit.

Since I (probably) had no better use of my time, I created quick script to record the average time of multiple webpack runs. It seemed to work good enough so I had it run webpack 10 times on one of my projects. These are the results!

**Current version on npm (1.0.0-beta.0):**
```
Test #1: 7226ms
Test #2: 6975ms
Test #3: 7251ms
Test #4: 7072ms
Test #5: 7311ms
Test #6: 7379ms
Test #7: 7190ms
Test #8: 7358ms
Test #9: 7038ms
Test #10: 7299ms

AVERAGE TIME: 7209.9ms
```

**This PR:**
```
Test #1: 8255ms
Test #2: 8263ms
Test #3: 8435ms
Test #4: 8642ms
Test #5: 8632ms
Test #6: 8555ms
Test #7: 8530ms
Test #8: 8544ms
Test #9: 8391ms
Test #10: 8518ms

AVERAGE TIME: 8476.5ms
```

As you can see, the average run time of this PR is `1266.6ms` slower than the current version on my project. This isn't that much in the grand scheme of things, but possibly something to consider.

<hr>

If anybody else wants to see what the performance is on their projects (this is only a sample size of 1 😄) - feel free to use my script. It's available here: https://gist.github.com/odensc/e1db530a9c37d142df6ed835ae75df5e

Save it to a file of your choice, and then run it with the command you would usually use for webpack, like so:
```bash
node time.js ./node_modules/.bin/webpack --config webpack.config.production.js
```

Hope this helps in some way!

michael-ciniawsky:
@odensc Thx 

There are 'discussion/plans'  to maybe adopt https://github.com/gdborton/webpack-parallel-uglify-plugin concepts, but a newer version of `uglify` needs to be supported and 'stable' first

Also a micro optimization is outstanding to move the `weakSet` into a different scope

d3viant0ne:
@odensc - Out of curiosity, what was the pre-minify size the bundles in your test project?

Curious if it's just a little slower across the board or if execution time is going to increase at an unacceptable rate as size increases.

Hartorn:
@michael-ciniawsky @hulkish @d3viant0ne Just trying to sum up guys : aside from documentation and performance consideration, the PR is alright ? Or I have things to change ?

I will try to make it parallel in an another branch, when I have some time. Will keep you up to date.

d3viant0ne:
Before this can land ...

1.) Tobias needs to have a look at it.
2.) We need to answer the performance question.

Outside of that, this is in pretty good shape now. I wouldn't expect many changes unless they come from Tobias.

odensc:
@d3viant0ne 1.37MB pre-minify total size.

graingert:
@d3viant0ne probably a bit unfair to compare uglify-js 2 with uglify-es 3. I think it would be good to compare all three.

d3viant0ne:
We aren't merging in all three. We are comparing the existing with the soon to be, it's the only performance increase or decrease that consumers are going to care about.

I'm not concerned with the performance of Uglify, I'm concerned with the performance of this plugin and what effect this pull request has on it.

mikesherov:
There might be something we could do here (I haven't looked deeply at this yet), but I wouldn't be surprised if this was just normal overhead from having to parse as es6 instead of es5.

d3viant0ne:
@mikesherov - That's totally possible and so long as the performance doesn't degrade exponentially as you add larger bundles, I personally feel the performance hit is acceptable unless there is some low handing perf fruit i'm not seeing.

Like most releases, the perf of ES6 / Uglify-ES should get better as time goes on. As long as we don't have a run away perf issue in either the plugin or ugilfy, i'm fine with this PR as is.

Just need to get Tobias to sign off on it and it's on to beta builds :)

michael-ciniawsky:
Fixes #32, #33, #66

jdalton:
I'd love to see this land. I'm currently using the last harmony tag of Uglify v2.x and being able to upgrade to v3 would be much appreciated.

michael-ciniawsky:
Any news from the core team on this ? That's the only thing missing to land it on master atm :)

graingert:
@michael-ciniawsky maybe @ tag them?

michael-ciniawsky:
I did already and ask in slack aswell no response until now ¯\_(ツ)_/¯

graingert:
Smash it in
Smash it in
Smash it in

michael-ciniawsky:
⚠️  Please wait with the next beta release I quickly finish the docs PR :)

d3viant0ne:
@jdalton @graingert ...

screen shot 2017-07-05 at 10 06 13 pm

d3viant0ne:
For all involved, this is staying on a beta tag at minimum until the test suite is reliable. I personally wouldn't have merged this yet given Tobias should have looked at it & we didn't answer the performance question though I do understand everyones desire to get this in master.

@Hartorn - Please see my comment in #67

openjck:
This hasn't been released yet, has it?

PR #67, opened 2017-07-06
ci(Travis): Enable OSX build on current master

d3viant0ne:
Don't Merge This, example for test timing issues currently in master

@Hartorn - If you don't mind, can you fix the es2015 test? There is a timing issue ( it's flakey ).

- The assertion fails randomly on Windows & OSX, see the build output of the second & third commit in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
-dist: trusty
+
+os:
+  - linux
+  - osx
+
 language: node_js
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 sudo: false
 
-os:
-  - linux
-  - osx
-
 language: node_js
 branches:
   only:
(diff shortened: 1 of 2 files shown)


Hartorn:
@d3viant0ne Sure, I will have a look. @hulkish put them in place, maybe he would be more relevant (basically I am a newbie considering unit testing, but will try my best).

d3viant0ne:
//cc @hulkish

hulkish:
ah... i wonder if this has to do with the error stack containing backslashes on windows in the snapshot? ill take a look.

d3viant0ne:
@hulkish - It's not a windows issue, that fails on OSX as well ( Note the expected vs. received )

screen_shot_2017-07-05_at_9 03 38_pm

hulkish:
@d3viant0ne hmm...im having a hard time replicating this behavior - I'm on macOS. Can you confirm you don't have any npm/yarn links that might be interfering?

Hartorn:
@d3viant0ne I tried to reproduce on Windows 10, both on git bash and cmd, I could not manage it, all the tests went green.
I will try some more this week-end, both on Linux Mint and Windows.

michael-ciniawsky:
Should this land or better enable macOS builds in gerenal via `webpack-defaults` ?

d3viant0ne:
@michael-ciniawsky - OSX builds block executions of any other build on Travis, thus we avoid it.

This really isn't an OSX issue, it was just the easiest way to highlight the issue with the test.

This pull request should be closed when no longer needed.

hulkish:
@d3viant0ne I think we're past this issue now, right? Can be closed?

PR #69, opened 2017-07-06
test: added tests to ensure correct uglify-es version is being used.

hulkish:
Fixes https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/67#issuecomment-313425627


d3viant0ne:
screen shot 2017-07-07 at 11 40 22 am

Attack of the modules .... again ! :p

hulkish:
Jeez, i guess i was incorrectly assuming the uglify-es version as the cause. what does this print out for u?

```bash
node -v
npm -v
cat node_modules/webpack/package.json | grep version
cat node_modules/uglify-es/package.json | grep version
cat node_modules/uglify-js/package.json | grep version
```

d3viant0ne:
screen shot 2017-07-07 at 12 05 17 pm

hulkish:
@d3viant0ne can u try installing npm >= 5? I suspect it's related to package-lock.json not being respected.

michael-ciniawsky:
It sometimes works locally (macOS) for me, but the test are definitely flaky. But what is this test actually testing here ? 🙃

hulkish:
@michael-ciniawsky the test in this PR is not actually intended to be merged. Just an attempt to find the cause of this issue.

PR #74, opened 2017-07-08
test: add test for `MultiCompiler` usage

### LINKED ISSUE #73: bug: uglifyjs-weback-plugin@1.0.0-beta.1 breaks with MultiCompiler
[This line](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/index.js#L152) breaks support for `webpack/lib/MultiCompiler` because a `MultiCompiler` instance child compiler does not have a `context`.

hulkish:
Fixes #73

**Update:** So, it looks like this issue just *magically fixed itself*. That, said - I'd like to still include the added tests and upgrade to `uglify-es@3.0.24`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,7 +149,10 @@ class UglifyJsPlugin {
   }
 
   apply(compiler) {
-    const requestShortener = new RequestShortener(compiler.context);
+    if (UglifyJsPlugin.appliedCompilers.has(compiler)) return;
+    UglifyJsPlugin.appliedCompilers.add(compiler);
+
+    const requestShortener = new RequestShortener(compiler.context || process.cwd());
     // Copy uglify options
     const uglifyOptions = UglifyJsPlugin.buildDefaultUglifyOptions(this.uglifyOptions);
     // Making sure output options exists if there is an extractComments options
@@ -279,5 +282,12 @@ class UglifyJsPlugin {
     });
   }
 }
+
+Object.defineProperty(UglifyJsPlugin, 'appliedCompilers', {
+  enumerable: false,
+  configurable: false,
+  writable: false,
+  value: new WeakSet(),
+});
 
 export default UglifyJsPlugin;
--- a/src/index.js
+++ b/src/index.js
@@ -149,9 +149,6 @@ class UglifyJsPlugin {
   }
 
   apply(compiler) {
-    if (UglifyJsPlugin.appliedCompilers.has(compiler)) return;
-    UglifyJsPlugin.appliedCompilers.add(compiler);
-
     const requestShortener = new RequestShortener(compiler.context || process.cwd());
     // Copy uglify options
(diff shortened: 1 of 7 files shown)


test/helpers.js
```diff
@@ -50,15 +50,29 @@ exports.createCompiler = function createCompiler(options = {}) {
...
+}
 
-exports.removeCWD = function removeCWD(str) {
+export function countPlugins({ _plugins }) {
+  return Object.keys(_plugins).reduce((aggregate, name) => {
+    const currentLength = typeof aggregate[name] === 'number' ? aggregate[name].length : 0;
+    const eventLength = Array.isArray(_plugins[name]) ? _plugins[name].length : 0;
+
+    if (eventLength > currentLength) {
+      Object.assign(aggregate, {
```
Kovensky: Why not just `aggregate[name] = eventLength`?

test/helpers.js
```diff
@@ -50,15 +50,29 @@ exports.createCompiler = function createCompiler(options = {}) {
   });
   compiler.outputFileSystem = new MemoryFileSystem();
   return compiler;
-};
+}
 
-exports.removeCWD = function removeCWD(str) {
+export function countPlugins({ _plugins }) {
```
Kovensky: Where is this used?

src/index.js
```diff
@@ -149,7 +149,7 @@ class UglifyJsPlugin {
   }
 
   apply(compiler) {
-    const requestShortener = new RequestShortener(compiler.context);
+    const requestShortener = new RequestShortener(compiler.context || process.cwd());
```
michael-ciniawsky: Is `process.cwd()` save in to use in most cases ?

test/helpers.js
michael-ciniawsky: What is the usecase here ? :)

src/index.js
hulkish: Oops, i have a regression test to add that uses this

d3viant0ne: @michael-ciniawsky - Agree here. We should get context without using `process. || window || w/e`

hulkish:
path.resolve()?

michael-ciniawsky:
> `path.resolve()` ?

For `compiler.context` you mean ? If can you post what the status quo is there atm please? I wonder why it isn't absolute and if this has a reason

hulkish:
@michael-ciniawsky yes, `path.resolve` & `process.cwd()` both seem to be used for this in webpack core: https://github.com/webpack/webpack/search?utf8=%E2%9C%93&q=%22new+RequestShortener%28%22&type=

test/helpers.js
hulkish: @Kovensky @michael-ciniawsky Was for the test I hadn't added at first to this pr. But, I have since added it: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/74/files#diff-0570d5214a7cbff246a7ad90d119f15c

PR #75, opened 2017-07-08
feat: ensure unique usage

### LINKED ISSUE #72: feature: Add logic to ensure only 1 UglifyJsPlugin's instance per compiler
## Problem:
Currently, I think it's too easy to accidentally apply 2 instances of UglifyJsPlugin to the same compiler/compilation.

### Scenario:
- You are a webpack newb
- You need to make a production build
- You learn UglifyJsPlugin is the recommended way to minify your bundle, so you make a `webpack.config.prod.js` and include UglifyJsPlugin.
- You learn that it's recommended to use the `--optimize-minimize/-p` argument with webpack cli, so u make a production build script:
```
"scripts": {
  "build:prod": "webpack -p"
}
```

Since this webpack cli option auto-adds the UglifyJsPlugin: 
- You now have attached 2 separate instances of the same plugin.
- Your production build is 'gon be sloowwwwww

## Solution
Add a guard to the top of `UglifyJsPlugin` class method `apply` which restricts multiple instances of the plugin from being applied to the same compiler, like so:

```js
class UglifyJsPlugin {
  static appliedCompilers = new WeakSet();

  // then use it as a guard:
  apply(compiler) {
    if (UglifyJsPlugin.appliedCompilers.has(compiler)) return;
    appliedCompilers.add(compiler);
    [...]
  }
}
```

### "But... that's user error"
Yes, it absolutely is. However, I have personally come across the mistake several times. My thinking is... the negative impact this potentially introduces is worth safeguarding against.

hulkish:
Closes #72

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,6 +149,7 @@ class UglifyJsPlugin {
   }
 
   apply(compiler) {
+    if (compiler.applyUnique(UglifyJsPlugin)) return;
     const requestShortener = new RequestShortener(compiler.context);
     // Copy uglify options
     const uglifyOptions = UglifyJsPlugin.buildDefaultUglifyOptions(this.uglifyOptions);
--- a/test/__snapshots__/unique-apply-per-compiler.test.js.snap
+++ b/test/__snapshots__/unique-apply-per-compiler.test.js.snap
@@ -1,0 +1,178 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`after adding a second plugin instance 1`] = `
+Object {
+  "after-emit": 1,
+  "after-resolvers": 2,
+  "before-run": 1,
+  "compilation": 32,
+  "entry-option": 1,
+  "make": 1,
+  "this-compilation": 1,
+}
+`;
+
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -280,4 +283,11 @@ class UglifyJsPlugin {
   }
 }
 
+Object.defineProperty(UglifyJsPlugin, 'appliedCompilers', {
```
michael-ciniawsky: Whats the difference between 
```js
Object.defineProperty(UglifyJSPlugin. 'appliedCompilers', {...})
```
 && 
```js
UglifyJSPlugin.appliedCompliers = new WeakSet()
```
?

test/__snapshots__/supports-multicompiler.test.js.snap
```diff
@@ -0,0 +1,116 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`compiler plugin count 1`] = `
+Object {
```
michael-ciniawsky: Is this the `compiler` `{Object}` ?

PR #77, opened 2017-07-09
feat: add support for `parallelization` && `caching` (`options.parallel`)

aui:
1. Use multiple processes to perform tasks in parallel.
2. Use the file caching system to skip files that have not changed.

Build time contrast:

```
| cache | workers | time(ms) |
| ------| ------- | -------- |
| false |       1 |    35216 |
| false |      31 |    16345 |
| true  |      31 |     3384 |
```
Demo: <https://travis-ci.org/aui/uglifyjs-webpack-plugin-demo>


michael-ciniawsky:
awesome 😊, but can you develop against  the `feature/webpack3` branch [aswell], which includes an upgrade to `uglify-es` (ES2015+ Minification) and is currently in `beta`   `npm i -D uglifyjs-webpack-plugin@beta` ?

src/index.js
```diff
@@ -34,23 +28,10 @@ class UglifyJsPlugin {
...
-      compress,
-      mangle: mangle == null ? true : mangle,
-      // Ignoring sourcemap from options
-      sourceMap: null,
-      output: { ...defaultUglifyOptions.output, ...output },
-      toplevel,
-      ie8,
-    };
+    this.options.uglifyOptions = this.options.uglifyOptions || {};
+    this.options.maxWorkers = this.options.maxWorkers || os.cpus().length;
```
```js
this.options.parallel = {
   cache: '{Boolean}' || 'path/to/cache',  
   workers: '{Number}'
}
```

`parallel`might not be the best name, open for suggestions here :)

src/uglify/cache.js
```diff
@@ -0,0 +1,20 @@
+import crypto from 'crypto';
+import cacache from 'cacache';
+
+const hashAlgorithm = 'sha512';
```
michael-ciniawsky: `hashAlgorithm` =>`hash || algorithm || sha512`

michael-ciniawsky:
[`compute-cluster`](https://www.npmjs.com/package/compute-cluster) seems to be fairly 'outdated' and is on `v0.0.9` with failing tests. Could you elaborate on your choice here please ? :). What about [`worker-farm`](https://www.npmjs.com/package/worker-farm) ?

hulkish:
i agree with @michael-ciniawsky, i really like `worker-farm` for this also.

package.json
```diff
@@ -30,8 +30,11 @@
     "webpack-defaults": "webpack-defaults"
   },
   "dependencies": {
+    "cacache": "^9.2.9",
+    "compute-cluster": "0.0.9",
+    "find-cache-dir": "^1.0.0",
     "source-map": "^0.5.6",
-    "uglify-es": "^3.0.21",
+    "uglify-es": "^3.0.24",
```
michael-ciniawsky: #74 merged, this should also update some of the snapshots you updated aswell :)

src/uglify/minify.js
```diff
@@ -0,0 +1,108 @@
...
+    mangle: mangle == null ? true : mangle,
+    // Ignoring sourcemap from options
+    sourceMap: null,
+    output: { ...defaultUglifyOptions.output, ...output },
+    toplevel,
+    ie8,
+  };
+};
+
+const buildCommentsFunction = (options, uglifyOptions, extractedComments) => {
```
michael-ciniawsky: `buildCommentsFunction` => `buildComments || comments`

src/uglify/minify.js
```diff
@@ -0,0 +1,108 @@
+import uglify from 'uglify-es';
+
+const defaultUglifyOptions = {
```
michael-ciniawsky: move the `{Object}` to L21 please :)

src/uglify/minify.js
```diff
@@ -0,0 +1,108 @@
...
+  output: {
+    comments: /^\**!|@preserve|@license|@cc_on/,
+    beautify: false,
+    semicolons: true,
+    shebang: true,
+  },
+};
+
+const buildDefaultUglifyOptions = ({ ecma, warnings, parse = {}, compress = {}, mangle, output, toplevel, ie8 }) => {
```
michael-ciniawsky: `buildDefaultUglifyOptions` => `buildOptions || defaults`

src/uglify/index.js
```diff
@@ -0,0 +1,73 @@
+import fs from 'fs';
+import path from 'path';
+import ComputeCluster from 'compute-cluster';
+import { get, put } from './cache';
+import { encode } from './serialization'; // eslint-disable-line import/newline-after-import
+const uglifyVersion = require(require.resolve('uglify-es/package.json')).version; // eslint-disable-line import/no-dynamic-require
```
michael-ciniawsky:
```js
const version = {
  uglify: require(require.resolve('uglify-es/package.json')).version,
  plugin: require('../../package.json').version // require.resolve() here aswell ?
}
```

Maybe consider a separate file `src/uglify/versions.js` ?

src/uglify/serialization.js
```diff
@@ -0,0 +1,46 @@
+const toType = value => (Object.prototype.toString.call(value).slice(8, -1));
```
michael-ciniawsky: Please give an example, why this is needed here

src/index.js
mikesherov: - 1 for cpus

michael-ciniawsky: @mikesherov Should this be a +1 for `options.cpus` ? 🙃

mikesherov: `length - 1`

mikesherov:
Great work here, a couple of notes:

1. Is workerization faster when CPU count is 1? Same question for when number of files to uglify is less than CPU count?
2. Separately, we should use worker-farm to manage processes.
3. Please max sure maxWorkers is maxed out at CPU count - 1. Having more workers than CPUs - 1 will introduce jank and slowdown.

src/index.js
```diff
@@ -166,13 +85,22 @@ class UglifyJsPlugin {
       }
 
       compilation.plugin('optimize-chunk-assets', (chunks, callback) => {
+        const uglify = new Uglify({
+          maxWorkers: this.options.maxWorkers,
+          cache: this.options.cache,
+          cacheDirectory: this.options.cacheDirectory,
+        });
+        const exit = () => {
+          uglify.exit(() => {
```
mikesherov: `const exit = () => uglify.exit(callback);`

package.json
```diff
@@ -12,9 +12,9 @@
     "start": "npm run build -- -w",
     "build": "cross-env NODE_ENV=production babel src -d dist --ignore 'src/**/*.test.js'",
     "test:all": "npm run test:coverage && npm run test:lint",
-    "test": "jest",
-    "test:coverage": "jest --collectCoverageFrom='src/**/*.js' --coverage",
-    "test:watch": "jest --watch",
+    "test": "jest --env node",
```
aui: Jest default configuration is `--env = jsdom`, it can simulate the browser environment, I worry that it takes up extra memory

d3viant0ne: Actually, Jest uses less memory than Karma does, by quite a bit. That said, this change if needed belongs in `webpack-defaults`.

All of our libs run the exact same projects setup. So for the scope of this pull request, this needs to be set back to what it was. 

If you feel that strongly about this particular configuration, open a pull request in https://github.com/webpack-contrib/webpack-defaults/issues so the change can be applied to all of our libs in the next patch release of defaults.

src/uglify/serialization.js
aui: Since `RegExp` or` Function` can not be serialized, they need special handling:

```js
worker.send({ extractComments: /foo/ });
```

```js
// worker
process.on('message', (message) => {
    console.log(message); // -> { extractComments: {} } 😢
})
```

Use serialization :

```js
worker.send(encode({ extractComments: /foo/ }));
```

```js
// worker
process.on('message', (message) => {
    message = decode(message);
    console.log(message); // -> { extractComments: /foo/ } 😊
})
```

test/uglify/serialization.test.js
```diff
@@ -0,0 +1,60 @@
...
+    expect(typeof json.func).toBe('function');
+    expect(json.func.toString()).toBe(input.func.toString());
+  });
+
+  it('serialization regexp', () => {
+    const input = {
+      regexp: /\s+/,
+    };
+    let json = encode(input);
+    expect(json.func).not.toBe(input.regexp);
```
mikesherov: `expect(json.regexp).not.toBe(input.regexp)`

test/uglify/serialization.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+  it('decode: deep copy of data', () => {
+    const json = decode(data);
+    expect(json).not.toBe(data);
+    expect(json.string).toBe(data.string);
+    expect(json.number).toBe(data.number);
+    expect(json.boolean).toBe(data.boolean);
+    expect(json.null).toBe(data.null);
+    expect(json.object.string).toBe(data.object.string);
+    expect(json.array.length).toBe(json.array.length);
```
mikesherov: `expect(json.array.length).toBe(data.array.length);`

test/uglify/serialization.test.js
```diff
@@ -0,0 +1,60 @@
...
+describe('serialization data', () => {
+  it('encode: deep copy of data', () => {
+    const json = encode(data);
+    expect(json).not.toBe(data);
+    expect(json.string).toBe(data.string);
+    expect(json.number).toBe(data.number);
+    expect(json.boolean).toBe(data.boolean);
+    expect(json.null).toBe(data.null);
+    expect(json.object.string).toBe(data.object.string);
+    expect(json.array.length).toBe(json.array.length);
```
mikesherov: `expect(json.array.length).toBe(data.array.length);`

src/uglify/serialization.js
mikesherov: I don't think we can legitimately do this for functions, though. 

```
var a = 1;
function toBeSerialized() {
  console.log(a);
}
```

in the case above, this results in non-obvious errors when serializing and deserializing in a different context. Is there any way we can avoid passing functions or regexes down to the worker and instead post process once comments are sent back to us?

src/uglify/index.js
```diff
@@ -0,0 +1,77 @@
...
+try {
+  const testWorkerFile = path.join(__dirname, '..', '..', 'dist', 'uglify', 'worker.js');
+  fs.accessSync(workerFile);
+  workerFile = testWorkerFile;
+} catch (e) { } // eslint-disable-line no-empty
+
+class Uglify {
+  constructor({
+    cache = findCacheDir({ name: 'uglifyjs-webpack-plugin' }),
+    workers = Math.max(os.cpus().length - 1, 1),
```
mikesherov: Hmm, this isn't quite right. If the user has  1 cpu, we want `workers` to be `0`, and use that as a signal to not fork any new processes, as that'll be much slower if we forked when there is only one CPU.

mikesherov: `workers = Math.max(os.cpus().length - 1, 1),` should be `workers = os.cpus().length - 1`, and we should take `0` to mean “in band”.

src/uglify/index.js
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import findCacheDir from 'find-cache-dir';
+import workerFarm from 'worker-farm';
+import { get, put } from './cache';
+import { encode } from './serialization';
+import versions from './versions';
+
+let workerFile = path.join(__dirname, 'worker.js');
```
hulkish: `path.join(__dirname, 'worker.js');` => `require.resolve('./worker')` ?

src/uglify/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+try {
+  const testWorkerFile = path.join(__dirname, '..', '..', 'dist', 'uglify', 'worker.js');
```
hulkish: `const testWorkerFile = path.join(__dirname, '..', '..', 'dist', 'uglify', 'worker.js');` => `require.resolve('../../dist/worker')` ?

michael-ciniawsky:
@aui We need to check for additional `cpu`  availability before entering 'worker mode' please, to avoid `seralization` overhead && eventual bugs with the `toType` check in the current `non-worker´ environment

src/index.js
```diff
@@ -182,6 +98,8 @@ class UglifyJsPlugin {
             try {
               let input;
               let inputSourceMap;
+              const dist = compiler.options ? compiler.options.output.path : '';
+              const id = path.join(dist, file);
```
hulkish: `path.join(dist, file);` => `path.resolve(dist, file);` ?

src/index.js
```diff
@@ -3,54 +3,31 @@
   Author Tobias Koppers @sokra
 */
 
+import path from 'path';
```
hulkish: This can be removed, If these are applied:

https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/77#pullrequestreview-49008895
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/77#pullrequestreview-49016047

src/uglify/serialization.js
aui: The Thank you for discovering this question I now think of three viable options:

1. When `parallel.workers> 0`, users are not allowed to use` Function` or` RegExp`
2. Before using serialization, use ESlint to check the "function" and report the "function" of context-dependent
3. At the run-time stage, catch the error and tell the user why the error occurred

I am currently inclined to the third program because it is friendly and simple. 😊

aui: @mikesherov I refactored the serialization.js, and now the user can know the details of the error :-)
```
Error: options serialization failed: "test" parse failed: 'console' is not defined.
    1| function toBeSerialized() {
 >> 2|     console.log(a);
    3|   }
```

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+    let toRun = tasks.length;
+    const results = [];
+    const step = (index, data) => {
+      toRun -= 1;
+      results[index] = data;
+      if (!toRun) {
+        callback(null, results);
+      }
+    };
+    if (!tasks.length) {
```
mikesherov: Move this guard all the way up

src/uglify/serialization.js
```diff
@@ -0,0 +1,121 @@
+/* eslint-disable
+  no-new-func
+*/
+import espree from 'espree';
```
mikesherov: The usage of espree and ESLint seems really really heavy handed here. I think we need to go back to the simpler approach of caching the error etc.

mikesherov: Yeah reread this whole file and while it's a neat trick, I think simply catching the error and warning that you can't use parallel if you provide your own functions is probably the best idea.

Or even providing a generic handler, and post processing the results so you don't have to serialize the regex or function.

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+try {
+  // run test
+  workerFile = require.resolve('../../dist/uglify/worker');
+} catch (e) { } // eslint-disable-line no-empty
+
+class Uglify {
+  constructor({
+    cache = findCacheDir({ name: 'uglifyjs-webpack-plugin' }),
+    workers = os.cpus().length - 1,
```
mikesherov: This still doesn't prevent a user from passing in a really high number of workers. Which I guess is ok?

mikesherov: Also, seems there's a better library from determining true number of CPUs available

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+    if (!tasks.length) {
+      process.nextTick(() => {
+        callback(null, results);
+      });
+      return;
+    }
+    tasks.forEach((task, index) => {
+      const cacheIdentifier = `${versions.uglify}|${versions.plugin}|${task.input}`;
+      const enqueue = () => {
+        this.worker(JSON.stringify(task, encode), (errors, data) => {
```
mikesherov: You shouldn't have to Stringify the task if it's not parallel, but I guess this is fine?

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+          const done = () => step(index, errors ? { error: errors.message } : data);
+          if (this.cache && !errors) {
+            put(this.cache, task.cacheKey, data, cacheIdentifier).then(done).catch(done);
+          } else {
+            done();
+          }
+        });
+      };
+      if (this.cache) {
+        get(this.cache, task.cacheKey, cacheIdentifier).then((data) => {
```
mikesherov: `.then(data => step(index, data), enqueue);`

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+import worker from './worker';
+
+let workerFile = require.resolve('./worker');
+
+try {
+  // run test
+  workerFile = require.resolve('../../dist/uglify/worker');
+} catch (e) { } // eslint-disable-line no-empty
+
+class Uglify {
```
mikesherov: Any reason you didn't just say `export default class Uglify`?

src/uglify/versions.js
```diff
@@ -0,0 +1,6 @@
+const version = {
```
mikesherov: Any reason to not do `export default const version = {`

src/uglify/index.js
```diff
@@ -0,0 +1,83 @@
...
+      });
+      return;
+    }
+    tasks.forEach((task, index) => {
+      const cacheIdentifier = `${versions.uglify}|${versions.plugin}|${task.input}`;
+      const enqueue = () => {
+        this.worker(JSON.stringify(task, encode), (errors, data) => {
+          const done = () => step(index, errors ? { error: errors.message } : data);
+          if (this.cache && !errors) {
+            put(this.cache, task.cacheKey, data, cacheIdentifier).then(done).catch(done);
```
mikesherov: .then(done, done)

src/uglify/versions.js
7rulnik: It's not valid JS. You can't export default variable declaration

mikesherov: Right sorry. Then just `export default {`

src/index.js
```diff
@@ -166,13 +80,13 @@ class UglifyJsPlugin {
       }
 
       compilation.plugin('optimize-chunk-assets', (chunks, callback) => {
+        const uglify = new Uglify(this.options.parallel);
```
michael-ciniawsky: `if (!this.options.parallel || (os.cpus().length - 1) === 0)` we need to use `uglify.minify()` as we do atm, without any worker related logic attached please :)

aui: Latest commit: `uglify/index.js` handles this logic, it keeps the plugin's original workflow :)

test/uglify/serialization.test.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  it('function operation failure should report critical information.', () => {
+    const a = 0;
+    const b = 1;
+    const input = {
+      func: () => a + b,
+    };
+    const string = JSON.stringify(input, encode);
+    expect(() => (JSON.parse(string, decode)).func()).toThrowError('the option "func" performs an error in the child process:');
```
michael-ciniawsky: `toThrowError(err)` => `toThrow(err)`

or/and maybe `expect(() => throws(err)).toThrowErrorMatchingSnapshot()`

michael-ciniawsky:
@aui Are there any options definitely known to not work in when `options.parallel` is enabled now ? I need to update the docs PR with the changes here accordingly, so anything thats needs attention in terms of documentation would be appreciated if you could give me a brief sum up on that 😛

test/empty-options.test.js
```diff
@@ -16,14 +16,14 @@ describe('when applied with no options', () => {
     const compilerEnv = pluginEnvironment.getEnvironmentStub();
     compilerEnv.context = '';
 
-    const plugin = new UglifyJsPlugin();
+    const plugin = new UglifyJsPlugin({ parallel: { cache: false, workers: 0 } });
```
hulkish: Please move all options related changes in this test to it's own test file, or to `parallel-option.test.js`. The purpose of this test is to literally have empty options.

test/es2015.test.js
```diff
@@ -17,6 +17,7 @@ describe('when applied with uglifyOptions.ecma', () => {
     });
 
     new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: Same comment I made for `empty-options.test`> It's important that these options for these tests are not changed in order for us to **really** understand that it is not breaking those use cases.

test/extract-option-set-to-a-single-file.test.js
```diff
@@ -11,6 +11,7 @@ describe('when applied with extract option set to a single file', () => {
     compilerEnv.context = '';
 
     const plugin = new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: Same comment here as with empty & es2015 options tests.

test/invalid-options.test.js
```diff
@@ -10,6 +10,7 @@ describe('when applied with invalid options', () => {
   it('matches snapshot', () => {
     const compiler = createCompiler();
     new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: Here too.

test/invalid-options.test.js
```diff
@@ -32,12 +33,13 @@ describe('when applied with invalid options', () => {
...
   });
 
-  it('outputs uglify errors', () => {
+  it('outputs uglify errors', (done) => {
     const pluginEnvironment = new PluginEnvironment();
     const compilerEnv = pluginEnvironment.getEnvironmentStub();
     compilerEnv.context = '';
 
     const plugin = new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: Here.

test/__snapshots__/invalid-options.test.js.snap
```diff
@@ -3,9 +3,9 @@
 exports[`errors 1`] = `
 Array [
   "Error: main.0c220ec66316af2c1b24.js from UglifyJs
-DefaultsError: \`invalid-option\` is not a supported option",
+\`invalid-option\` is not a supported option",
```
hulkish: Where id `DefaultsError` go?

test/__snapshots__/invalid-options.test.js.snap
```diff
@@ -3,9 +3,9 @@
...
   "Error: manifest.6afe1bc6685e9ab36c1c.js from UglifyJs
-DefaultsError: \`invalid-option\` is not a supported option",
+\`invalid-option\` is not a supported option",
```
hulkish: Again about `DefaultsError`

test/all-options.test.js
```diff
@@ -17,6 +17,7 @@ describe('when applied with all options', () => {
     compilerEnv.context = '';
 
     const plugin = new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: This one's fine, since this is an `all-options` test.

test/all-options.test.js
```diff
@@ -311,13 +325,14 @@ describe('when applied with all options', () => {
               compilationEventBindings = chunkPluginEnvironment.getEventBindings();
             });
 
-            it('should get all warnings', () => {
+            it('should get all warnings', (done) => {
```
hulkish: Remove, because this is a synchronous test (ensured via `PluginEnvironment`). This should not use a callback. or return a promise. This will also help to make sure the `parallel` feature being added does not break this expected behavior.

test/all-options.test.js
```diff
@@ -365,28 +380,16 @@ describe('when applied with all options', () => {
               compilationEventBindings = chunkPluginEnvironment.getEventBindings();
             });
 
-            it('should get no warnings', () => {
+            it('should get no warnings', (done) => {
```
hulkish: remove

test/all-options.test.js
```diff
@@ -365,28 +380,16 @@ describe('when applied with all options', () => {
...
               compilationEventBindings[1].handler([{
                 files: ['test2.js'],
               }], () => {
                 expect(compilation.warnings.length).toBe(0);
+                done();
```
hulkish: remove

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
-          expect(callback.mock.calls.length).toBe(1);
+          compilationEventBinding.handler([''], () => {
+            callback();
+            expect(callback.mock.calls.length).toBe(1);
+            done();
+          });
         });
 
-        it('default only parses filenames ending with .js', () => {
+        it('default only parses filenames ending with .js', (done) => {
```
hulkish: remove

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test', 'test.js'],
           }], () => {
             expect(Object.keys(compilation.assets).length).toBe(4);
+            done();
```
hulkish: remove

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('early returns if private property is already set', () => {
+        it('early returns if private property is already set', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test.js'],
           }], () => {
             expect(compilation.assets['test.js']).toEqual({});
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs stack trace errors for invalid asset', () => {
+        it('outputs stack trace errors for invalid asset', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test1.js'],
           }], () => {
             expect(compilation.errors.length).toBe(1);
             expect(compilation.errors[0]).toBeInstanceOf(Error);
-            expect(compilation.errors[0].message).toEqual(expect.stringContaining('TypeError'));
+            expect(compilation.errors[0].message).toEqual(expect.stringContaining('asset.source is not a function'));
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs parsing errors for invalid javascript', () => {
+        it('outputs parsing errors for invalid javascript', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test2.js'],
           }], () => {
             expect(compilation.errors.length).toBe(1);
             expect(compilation.errors[0]).toBeInstanceOf(Error);
             expect(compilation.errors[0].message).toEqual(expect.stringContaining('Unexpected token'));
             expect(compilation.errors[0].message).toEqual(expect.stringContaining('[test2.js:1,8]'));
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs no errors for valid javascript', () => {
+        it('outputs no errors for valid javascript', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs RawSource for valid javascript', () => {
+        it('outputs RawSource for valid javascript', (done) => {
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             expect(compilation.assets['test3.js']).toBeInstanceOf(RawSource);
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs no errors for valid javascript', () => {
+        it('outputs no errors for valid javascript', (done) => {
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             expect(compilation.errors.length).toBe(0);
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('outputs RawSource for valid javascript', () => {
+        it('outputs RawSource for valid javascript', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             expect(compilation.assets['test3.js']).toBeInstanceOf(RawSource);
+            done();
           });
         });
 
-        it('outputs mangled javascript', () => {
+        it('outputs mangled javascript', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             // eslint-disable-next-line no-underscore-dangle
             expect(compilation.assets['test3.js']._value)
               .not.toEqual(expect.stringContaining('longVariableName'));
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('compresses and does not output beautified javascript', () => {
+        it('compresses and does not output beautified javascript', (done) => {
```
hulkish: remove the `done`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             // eslint-disable-next-line no-underscore-dangle
             expect(compilation.assets['test3.js']._value).not.toEqual(expect.stringContaining('\n'));
+            done();
```
hulkish: remove the `done();`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           });
         });
 
-        it('preserves comments', () => {
+        it('preserves comments', (done) => {
```
hulkish: remove the `done(`

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
...
           compilationEventBinding.handler([{
             files: ['test3.js'],
           }], () => {
             // eslint-disable-next-line no-underscore-dangle
             expect(compilation.assets['test3.js']._value).toEqual(expect.stringContaining('/**'));
+            done();
```
hulkish: remove the `done();`

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
     [compilationEventBinding] = chunkPluginEnvironment.getEventBindings();
   });
 
-  it('normalizes extractConmments', () => {
+  it('normalizes extractConmments', (done) => {
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
...
     compilationEventBinding.handler([{
       files: ['test.js'],
     }], () => {
       expect(compilation.errors.length).toBe(0);
+      done();
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
...
     });
   });
 
-  it('outputs warnings for unreachable code', () => {
+  it('outputs warnings for unreachable code', (done) => {
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
...
     compilationEventBinding.handler([{
       files: ['test.js', 'test1.js'],
     }], () => {
       expect(compilation.warnings.length).toBe(1);
       expect(compilation.warnings[0]).toBeInstanceOf(Error);
       expect(compilation.warnings[0].message).toEqual(expect.stringContaining('Dropping unreachable code'));
+      done();
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
...
     });
   });
 
-  describe('normalizes when options.extractComments is regex', () => {
+  it('normalizes when options.extractComments is regex', (done) => {
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -102,15 +106,17 @@ describe('when options.extractComments', () => {
       files: ['test.js'],
     }], () => {
       expect(compilation2.errors.length).toBe(0);
+      done();
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -102,15 +106,17 @@ describe('when options.extractComments', () => {
...
     });
   });
 
-  describe('converts boolean options.extractComments.condition to function', () => {
+  it('converts boolean options.extractComments.condition to function', (done) => {
```
hulkish: remove

test/extract-comments-options.test.js
```diff
@@ -53,30 +54,33 @@ describe('when options.extractComments', () => {
...
   });
 
-  describe('normalizes when options.extractComments is regex', () => {
+  it('normalizes when options.extractComments is regex', (done) => {
     const pluginEnvironment = new PluginEnvironment();
     const compilerEnv = pluginEnvironment.getEnvironmentStub();
     compilerEnv.context = '';
 
     const plugin = new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: This one is ok, because it's helpful to test this against es2015 feature.

test/extract-comments-options.test.js
```diff
@@ -102,15 +106,17 @@ describe('when options.extractComments', () => {
...
   });
 
-  describe('converts boolean options.extractComments.condition to function', () => {
+  it('converts boolean options.extractComments.condition to function', (done) => {
     const pluginEnvironment = new PluginEnvironment();
     const compilerEnv = pluginEnvironment.getEnvironmentStub();
     compilerEnv.context = '';
 
     const plugin = new UglifyJsPlugin({
+      parallel: { cache: false, workers: 0 },
```
hulkish: This one is ok

test/extract-comments-options.test.js
```diff
@@ -144,6 +150,7 @@ describe('when options.extractComments', () => {
       files: ['test.js'],
     }], () => {
       expect(compilation2.errors.length).toBe(0);
+      done();
```
hulkish: remove

test/extract-option-set-to-a-single-file.test.js
```diff
@@ -74,18 +75,19 @@ describe('when applied with extract option set to a single file', () => {
           [compilationEventBinding] = compilationEventBindings;
         });
 
-        it('preserves comments', () => {
+        it('preserves comments', (done) => {
```
hulkish: remove

test/extract-option-set-to-a-single-file.test.js
```diff
@@ -74,18 +75,19 @@ describe('when applied with extract option set to a single file', () => {
...
           compilationEventBinding.handler([{
             files: ['test.js', 'test2.js', 'test3.js'],
           }], () => {
             expect(compilation.assets['test.js'].source()).toEqual(expect.stringContaining('/*'));
             expect(compilation.assets['test2.js'].source()).toEqual(expect.stringContaining('//'));
             expect(compilation.assets['test3.js'].source()).toEqual(expect.stringContaining('/*'));
             expect(compilation.assets['test3.js'].source()).toEqual(expect.stringContaining('//'));
+            done();
```
hulkish: remove

test/extract-option-set-to-a-single-file.test.js
```diff
@@ -74,18 +75,19 @@ describe('when applied with extract option set to a single file', () => {
...
           });
         });
 
-        it('extracts comments to specified file', () => {
+        it('extracts comments to specified file', (done) => {
```
hulkish: remove

test/extract-option-set-to-a-single-file.test.js
```diff
@@ -95,6 +97,7 @@ describe('when applied with extract option set to a single file', () => {
             expect(compilation.assets['extracted-comments.js'].source()).toEqual(expect.stringContaining('/* This is a comment from test3.js */'));
             expect(compilation.assets['extracted-comments.js'].source()).toEqual(expect.stringContaining('// This is another comment from test3.js'));
             expect(compilation.assets['extracted-comments.js'].source()).not.toEqual(expect.stringContaining('function'));
+            done();
```
hulkish: remove

test/invalid-options.test.js
```diff
@@ -32,12 +33,13 @@ describe('when applied with invalid options', () => {
     });
   });
 
-  it('outputs uglify errors', () => {
+  it('outputs uglify errors', (done) => {
```
hulkish: remove

test/invalid-options.test.js
```diff
@@ -65,6 +67,7 @@ describe('when applied with invalid options', () => {
       expect(compilation.errors.length).toBe(1);
       expect(compilation.errors[0]).toBeInstanceOf(Error);
       expect(compilation.errors[0].message).toEqual(expect.stringContaining('from UglifyJs'));
+      done();
```
hulkish: remove

test/empty-options.test.js
```diff
@@ -93,90 +93,102 @@ describe('when applied with no options', () => {
           expect(compilationEventBinding.name).toEqual('optimize-chunk-assets');
         });
 
-        it('only calls callback once', () => {
+        it('only calls callback once', (done) => {
```
hulkish: remove

test/empty-options.test.js
```diff
@@ -16,14 +16,14 @@ describe('when applied with no options', () => {
...
-    const plugin = new UglifyJsPlugin();
+    const plugin = new UglifyJsPlugin({ parallel: { cache: false, workers: 0 } });
     plugin.apply(compilerEnv);
     eventBindings = pluginEnvironment.getEventBindings();
   });
 
   it('matches snapshot', () => {
     const compiler = createCompiler();
-    new UglifyJsPlugin().apply(compiler);
+    new UglifyJsPlugin({ parallel: { cache: false } }).apply(compiler);
```
hulkish: Please remove this change here and just make a new test inside `parallel-option.test.ja`. Something like:
  it('matches snapshot', () => {
    const compiler = createCompiler();
    new UglifyJsPlugin({ parallel: { cache: false } }).apply(compiler);

    return compile(compiler).then((stats) => {
      const errors = stats.compilation.errors.map(cleanErrorStack);
      const warnings = stats.compilation.warnings.map(cleanErrorStack);

      expect(errors).toMatchSnapshot('errors');
      expect(warnings).toMatchSnapshot('warnings');

      for (const file in stats.compilation.assets) {
        if (Object.prototype.hasOwnProperty.call(stats.compilation.assets, file)) {
          expect(stats.compilation.assets[file].source()).toMatchSnapshot(file);
        }
      }
    });
  });

test/__snapshots__/invalid-options.test.js.snap
aui: I fixed the BUG: `src / index.js> buildError ()`. So the results are not the same as before:

```js
    } else if (err.msg) {
      return new Error(`${file} from UglifyJs\n${err.msg}`);
    }
    return new Error(`${file} from UglifyJs\n${err.stack}`); // -> ...DefaultsError: `invalid-option` is not a supported option"
```

```js
    } else if (err.message) {
      return new Error(`${file} from UglifyJs\n${err.message}`); // -> ...`invalid-option` is not a supported option"
    }
    return new Error(`${file} from UglifyJs\n${err.stack}`);
```

test/__snapshots__/invalid-options.test.js.snap
aui: `uglify-es` error object does not have a `msg` field.

src/index.js \> buildError() :

```js
} else if (err.msg) {
  return new Error(`${file} from UglifyJs\n${err.msg}`);
}
return new Error(`${file} from UglifyJs\n${err.stack}`); // -> ...DefaultsError: `invalid-option` is not a supported option"
```

```js
} else if (err.message) {
  return new Error(`${file} from UglifyJs\n${err.message}`); // -> ...`invalid-option` is not a supported option"
}
return new Error(`${file} from UglifyJs\n${err.stack}`);
```

hulkish: can you try changing `(this line)[https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/test/helpers.js#L68] to: 

```js
const str = exports.removeCWD(error.message)
```

test/empty-options.test.js
```diff
@@ -121,7 +123,7 @@ describe('when applied with no options', () => {
           }], () => {
             expect(compilation.errors.length).toBe(1);
             expect(compilation.errors[0]).toBeInstanceOf(Error);
-            expect(compilation.errors[0].message).toEqual(expect.stringContaining('TypeError'));
```
hulkish: Just making note of this for when #80 is rebased on top of this. This should remain a TypeError - but I think the options schema validation should cover this.

mikesherov:
Wooo!

scott-thrillist:
👍  Waiting for this to get on npm to compare with `webpack-parallel-uglify-plugin` ;)

PR #64, opened 2017-06-30
feat(index): add `options` validation

michael-ciniawsky:

PR #80, opened 2017-07-12
feat(index): add `options` validation (`schema-utils`)

### LINKED ISSUE #4: feat: add options validation
It should be processed through [ajv](https://www.npmjs.com/package/ajv) to give proper errors if the schema validation fails.

hulkish:
Closes #4, https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/64.


src/index.js
```diff
@@ -150,11 +140,11 @@ class UglifyJsPlugin {
...
     const requestShortener = new RequestShortener(compiler.context);
-    // Copy uglify options
-    const uglifyOptions = UglifyJsPlugin.buildDefaultUglifyOptions(this.uglifyOptions);
-    // Making sure output options exists if there is an extractComments options
-    if (this.options.extractComments) {
-      uglifyOptions.output = uglifyOptions.output || {};
+    const { uglifyOptions } = this.options;
+    const uglifiedAssets = new WeakSet();
+
+    if (typeof this.options.sourceMap === 'undefined' && (compiler.options.devtool === 'sourcemap' || compiler.options.devtool === 'source-map')) {
```
evilebottnawi: Seems not related to this PR

hulkish: You're right. I can separate this into a diff pr if u like. 

However, I gave this a good amount of thought and I believe it's been a missing piece to this plugin for some time now.

Not to mention - [this same logic is built into webpack core](https://github.com/webpack/webpack/blob/0ab36c491ea547add35af0c3fbbcdfab03a4c15e/bin/convert-argv.js#L441). I believe this responsibility belongs in this plugin.

If this pr is merged i plan to follow up by making a pr at webpack core repo to extract this logic from there.

evilebottnawi: @hulkish let's wait what other will say, but be good have separate PR for this, it is allow to maintenance CHANGELOG

hulkish: ok sure np

evilebottnawi: > If this pr is merged i plan to follow up by making a pr at webpack core repo to extract this logic from there. This also doesn't introduce a breaking change and all tests are passing.

btw good enchantment!

hulkish: bahaha : 🌮

src/index.js
```diff
@@ -7,49 +7,39 @@ import { SourceMapConsumer } from 'source-map';
...
 /* eslint-disable
   no-param-reassign
 */
 
 const warningRegex = /\[.+:([0-9]+),([0-9]+)\]/;
 
-const defaultUglifyOptions = {
-  output: {
-    comments: /^\**!|@preserve|@license|@cc_on/,
-    beautify: false,
```
hulkish: Removed these because these are the same defaults `uglify-es` uses internallu.

package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "start": "npm run build -- -w",
-    "build": "cross-env NODE_ENV=production babel src -d dist --ignore 'src/**/*.test.js'",
```
hulkish: added this because babel-cli to auto-copy src/options-schema.json to the dist folder. (to prevent MODULE_NOT_FOUND error on published code)

src/index.js
```diff
@@ -7,49 +7,39 @@ import { SourceMapConsumer } from 'source-map';
...
+      test = /\.js$/i,
+      warningsFilter = () => true,
+      extractComments,
+      sourceMap,
+    } = options;
 
-    this.uglifyOptions = this.options.uglifyOptions || {};
-  }
-
-  static buildDefaultUglifyOptions({ ecma, warnings, parse = {}, compress = {}, mangle, output, toplevel, ie8 }) {
```
hulkish: Simplified this by leveraging destructing default values.

src/index.js
```diff
@@ -166,7 +156,6 @@ class UglifyJsPlugin {
       }
 
       compilation.plugin('optimize-chunk-assets', (chunks, callback) => {
-        const uglifiedAssets = new WeakSet();
```
hulkish: Hoisted this to `apply` scope. This was my change originally and was an optimization that I wanted to include but was late to do so before it got merged. I've tested it pretty thoroughly.

src/options-schema.json
```diff
@@ -0,0 +1,50 @@
+{
```
hulkish: Reworked from #64

src/index.js
```diff
@@ -150,11 +140,11 @@ class UglifyJsPlugin {
 
   apply(compiler) {
     const requestShortener = new RequestShortener(compiler.context);
-    // Copy uglify options
-    const uglifyOptions = UglifyJsPlugin.buildDefaultUglifyOptions(this.uglifyOptions);
-    // Making sure output options exists if there is an extractComments options
-    if (this.options.extractComments) {
-      uglifyOptions.output = uglifyOptions.output || {};
```
hulkish: replaced with logic in constructor

src/index.js
```diff
@@ -7,49 +7,39 @@ import { SourceMapConsumer } from 'source-map';
 import { SourceMapSource, RawSource, ConcatSource } from 'webpack-sources';
 import RequestShortener from 'webpack/lib/RequestShortener';
 import ModuleFilenameHelpers from 'webpack/lib/ModuleFilenameHelpers';
+import validateOptions from 'schema-utils';
 import uglify from 'uglify-es';
+import optionsSchema from './options-schema.json';
```
michael-ciniawsky: `optionsSchema` => `schema`

src/index.js
```diff
@@ -7,49 +7,39 @@ import { SourceMapConsumer } from 'source-map';
...
-
 class UglifyJsPlugin {
-  constructor(options) {
-    if (typeof options !== 'object' || Array.isArray(options)) {
-      this.options = {};
-    } else {
-      this.options = options || {};
-    }
+  constructor(options = {}) {
+    validateOptions(optionsSchema, options, 'UglifyJsPlugin');
```
michael-ciniawsky: `UglifyJsPlugin` => `UglifyJS Plugin`

src/options-schema.json
michael-ciniawsky: `src/options-schema.json` => `./options.json`, if  it should reside in the src and needs to be copied, the change needs to be made in  `webpack-defaults` beforehand as always 😛

src/options-schema.json
```diff
@@ -0,0 +1,50 @@
...
+        "mangle": {
+          "type": ["boolean", "object"],
+          "properties": {
+            "properties": {
+              "type": ["boolean", "object"]
+            }
+          }
+        },
+        "output": {
+          "type": ["object", "null"]
```
michael-ciniawsky: Is `null` valid ? 🙃

src/options-schema.json
```diff
@@ -0,0 +1,50 @@
...
+    "sourceMap": {
+      "type": "boolean"
+    },
+    "uglifyOptions": {
+      "additionalProperties": true,
+      "type": "object",
+      "properties": {
+        "ecma": {
+          "type": "number",
+          "enum": [5, 6, 7, 8]
```
michael-ciniawsky: Could make this throw and see if the `ValidationError` is still displayed correctly with this additional info please?

package.json
michael-ciniawsky: => `webpack-defaults`, but `options.json` in the root is also ok imho,  since it is metadata, still needs to be added to `pkg.files` then, so really no preference :)

src/options-schema.json
hulkish: yes, it's default value is null: https://github.com/mishoo/UglifyJS2#user-content-minify-options

michael-ciniawsky: I was referring to the schema validation here :) e.g `{Regex}, {Function}` aren't supported  by `ajv` (`schema-utils`) yet.

hulkish:
ok done @michael-ciniawsky https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/80/commits/2a3687e9de2fca2d22cd0366226da9b893020d8e

options.json
```diff
@@ -0,0 +1,51 @@
...
+    "test": {},
+    "warningsFilter": {},
+    "extractComments": {},
+    "sourceMap": {
+      "type": "boolean"
+    },
+    "uglifyOptions": {
+      "additionalProperties": true,
+      "type": "object",
+      "properties": {
```
michael-ciniawsky: `ie8` ?

options.json
```diff
@@ -0,0 +1,51 @@
+{
+  "type": "object",
+  "properties": {
```
michael-ciniawsky: `include, exclude, parallel` ?

package.json
```diff
@@ -6,7 +6,8 @@
   "license": "MIT",
   "main": "dist/cjs.js",
   "files": [
-    "dist"
+    "dist",
+    "options.json"
```
michael-ciniawsky: I'm 💯 sry about this, your initial way of handling this file via `--copy-files` in the `build` script is the better solution, please revert :) 👍

src/index.js
```diff
@@ -7,6 +7,8 @@ import { SourceMapConsumer } from 'source-map';
 import { SourceMapSource, RawSource, ConcatSource } from 'webpack-sources';
 import RequestShortener from 'webpack/lib/RequestShortener';
 import ModuleFilenameHelpers from 'webpack/lib/ModuleFilenameHelpers';
+import validateOptions from 'schema-utils';
+import schema from '../options.json';
```
michael-ciniawsky: `../options.json` => `./options.json` (`../` => `./`) ?

hulkish: yea, I thought that you wanted me to include that with the --copy-files change.. ill switch it back

PR #81, opened 2017-07-12
feat: support `devtool` (`options.sourceMap`)

hulkish:
As described here: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/80#discussion_r126919798


test/all-options.test.js
```diff
@@ -82,6 +82,60 @@ describe('when applied with all options', () => {
...
+
+      return Promise.all([
+        compile(compiler1).then((stats) => {
+          const errors = stats.compilation.errors.map(cleanErrorStack);
+          const warnings = stats.compilation.warnings.map(cleanErrorStack);
+
+          expect(errors).toMatchSnapshot('devtool === "sourcemap" errors');
+          expect(warnings).toMatchSnapshot('devtool === "sourcemap" warnings');
+
+          for (const file in stats.compilation.assets) {
```
michael-ciniawsky:
```js
const assets = stats.compilation.assets

Object.keys(assets).forEach((asset) => {
  const source = assets[asset].source()
  return expect(source).toMatchSnapshot(asset)
})
```

src/index.js
```diff
@@ -157,6 +157,10 @@ class UglifyJsPlugin {
       uglifyOptions.output = uglifyOptions.output || {};
     }
 
+    if (typeof this.options.sourceMap === 'undefined' && ['sourcemap', 'source-map', 'hidden-source-map'].indexOf(compiler.options.devtool) > -1) {
```
michael-ciniawsky: Is the `typeof this.options.sourceMap === 'undefined'` necessary here ?

hulkish: Yes, if we're talking current usage. Because it can be user defined as true or false. This basically only auto-sets the sourceMap option value if u arent manually overriding it

test/all-options.test.js
hulkish: will do..but i dont think we want we `return` in that `forEach`

michael-ciniawsky: Yup we don't want to`return` there , updated 😛

src/index.js
michael-ciniawsky: If `devtool` should be automatically supported, the previous `this.options.sourceMap`  'doesn't matter'  as it is overridden to `true` anyways in case `compilation.options.devtool` is set ? Maybe I'm missing a case here 🙃

michael-ciniawsky: And if I missed one `if (typeof this.options.sourceMap === "undefined")` => `if (!this.options.sourceMap)` ?

hulkish: well, honestly I'm not certain of the history of this option.. I was just being careful to not break it's current usage.

That said, i see your point and im alright with removing the exposing of this option all together. Up to u.

hulkish: @michael-ciniawsky what u think?

mikesherov:
My vote is this logic should remain in webpack proper, as the logic has even changed since this PR was suggested. Just my 2 cents.

michael-ciniawsky:
Agreeing with @mikesherov here, this adds unnecessary side effects to the plugin

PR #106, opened 2017-08-07
Restore deleted documentation

alcuadrado:
Hi,

PR #61 deleted this sentence from the README which was really helpful. I found it by chance looking at tag v4.6's README.md while trying to use uglify-es with webpack. I suspect it was removed by accident, as it stated very clearly why one would use this package.


michael-ciniawsky:
The README on master refers to to the current `beta` and there is also a link to the `v0.4.6` docs. This info doesn't apply for `v1.0.0` and is incorrect for `>= v0.4.6` aswell, which ships with a fixed version of `uglify-js` to avoid breaking the webpack build

alcuadrado:
Ohh, I see! I misunderstood the comment in `v0.4.6`. Sorry for the inconvenience!

Issue #100, opened 2017-07-31
[v1.0.0-beta.2] cacheKey doesn't take uglifyOptions into account


mikesherov:
The cache key for cached version of the output should also include the options passed to uglify, otherwise, the same input with different options will yeild incorrect output.

hulkish:
@mikesherov What solution do u recommend?

mikesherov:
No, I recommend stringifying the uglifyOptions as part of the cacheKey generation at the same spot uglify version number is appended.

Issue #97, opened 2017-07-29
[v1.0.0-beta.2] parallelization isn't working


stevefan1999:
Well I was hacking next.js and tried to play with the new toy in town

I attempted to remove the old uglifyJs from webpack and installed the beta version back into the plugins config in wepback:
```js
removeUglifyJsPlugin () {
  // create a clone of config.plugins without UglifyJsPlugin
  delete this.config.plugins.filter(p => p.constructor.name === 'UglifyJsPlugin')[0]
  this.config.plugins = this.config.plugins.filter(p => p.constructor.name !== 'UglifyJsPlugin')

  return this
}

addUglifyEsPlugin (opts) {
  const plugin = new UglifyJSPlugin({
    parallel: true,
    sourceMap: opts.sourceMap,
    uglifyOptions: { // UglifyJS config for every workers
      ...opts, 
      ecma: 5
    }
  })

  this.config.plugins.push(plugin)
  return this
}
```
So then I built several times
```
$ next build
> Using external babel configuration
> Location: "/mnt/d/Seafile/power-next/.babelrc"
> Using "webpack" config function defined in next.config.js.
Done in 23.60s.

$ next build
> Using external babel configuration
> Location: "/mnt/d/Seafile/power-next/.babelrc"
> Using "webpack" config function defined in next.config.js.
Done in 24.89s.

$ next build
> Using external babel configuration
> Location: "/mnt/d/Seafile/power-next/.babelrc"
> Using "webpack" config function defined in next.config.js.
Done in 21.28s.
```
~22s, this is pretty slow...
I can validate uglify-es is working by checking the output(which is originally ES6+)

So I decided to turn off parallelization:
```js
addUglifyEsPlugin (opts) {
  const plugin = new UglifyJSPlugin({
    parallel: false,
    sourceMap: opts.sourceMap,
    uglifyOptions: { // UglifyJS config for every workers
      ...opts, 
      ecma: 5
    }
  })

  this.config.plugins.push(plugin)
  return this
}
```
Shockingly the build time is almost the same with parallel option off!
```
$ next build
> Using external babel configuration
> Location: "/mnt/d/Seafile/power-next/.babelrc"
> Using "webpack" config function defined in next.config.js.
Done in 22.37s.

$ next build
> Using external babel configuration
> Location: "/mnt/d/Seafile/power-next/.babelrc"
> Using "webpack" config function defined in next.config.js.
Done in 22.63s.

$ next build
[...]

PR #108, opened 2017-08-08
refactor: split `cache` and `parallel` options

evilebottnawi:
### Notable Changes

1. Separate `cache` and `parallel` options, because they should not be tied up. I can enable `cache` but disable `parallel`, also i can enable `parallel` but disable `cache`. Seems we were not too careful when we were accepted PR with `parallel` option.
2. Fix https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/97, options `parallel` doesn't work as expected (https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/97).
3. A lot of tests (`cache`, `parallel`, `ecma-8`, `ecma-7`, `ecma-6` `ecma-5`). Remove unnecessary extra tests from `parallel` tests.
4. Refactor some properties and method to easy read.
5. Simplify: remove `src/uglify/cache.js`, use `cacache` directly instead.
6. Don't use `cacheIdentifier` (aka `metadata`) field for invalidate cache because `metadata` should be used for other purposes (example `metadata` https://github.com/zkat/cacache#example-2). For invalidate cache we should use `cacheKey`.
7. Invalidate cache when `uglify-es` was updated.
8. Invalidate cache when `uglifyjs-webpack-plugin` was updated.
9. Invalidate cache when `options` for plugin was changed (we can't use only `uglifyOptions` because we can change `extractComments` or `warningsFilter` options). Maybe we can exclude some options not affecting on cache (`test`, `include`, `exclude`, `parallel` and `sourceMap`). Awaiting feedback.

### Issues

Fix https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/100
Fix https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/97

### One strange behavior
I remove `[chunkhash]` in `ecma.test.js` because `webpack` generate difference `[chunkhash]` for `ecma-5` on node `4` and `8`, I don't know why :/


src/uglify/index.js
```diff
@@ -1,10 +1,9 @@
 import os from 'os';
+import cacache from 'cacache';
 import findCacheDir from 'find-cache-dir';
```
michael-ciniawsky: `findCacheDir` => `cacheDir`

src/uglify/index.js
```diff
@@ -1,10 +1,9 @@
 import os from 'os';
+import cacache from 'cacache';
```
michael-ciniawsky: `cacache` => `cache`

test/uglify/cache.test.js
```diff
@@ -1,30 +0,0 @@
-import crypto from 'crypto';
```
michael-ciniawsky: Is the caching sufficiently tested elsewhere in the tests ?

evilebottnawi: @michael-ciniawsky be good add more tests, but `cache.test.js` doesn't test caching (just test `get` and `put` function)

evilebottnawi: let's add more tests

src/uglify/index.js
evilebottnawi: @michael-ciniawsky i think rename can be misleading, we can use `cache` variable for other purpose. Also i think it is bad practice, but if someone from the reviewers considers renaming a good practice, say it here and I do it :smile:

evilebottnawi:
@michael-ciniawsky @hulkish @bebraw @d3viant0ne friendly ping
Sorry for many changes, but current version don't work as expected :disappointed: 
After merge this PR i will do:
1. Refactor tests (because now we got `Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.` if something unexpected in tests, but should get pretty output about errors)
2. Implement `include` and `exclude` options.

michael-ciniawsky:
> Refactor tests (now we got Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. error if something unexpected in tests, but should get pretty output about errors)

This happend for me locally when I didn't build the `src` before running the tests (shouldn't be needed)

evilebottnawi:
@michael-ciniawsky we have `beta` and can change this :smile:

src/uglify/index.js
evilebottnawi: @Kovensky @michael-ciniawsky let's rename, thanks!

michael-ciniawsky:
Yeah it's not optimal, but the `cache` as standalone option definitely makes more sense, let's break early until we 'can' 😛 . I review in depth later today

src/uglify/index.js
michael-ciniawsky: @Kovensky It is a `{Function}` which 'contains' the value =>

1. Directly pass as an argument with 'higher order' => `const cache = (data, cacheDir(options)) => cache`
2. Assign it to a scope/block/'namespace' instead of an 'redundant' variable => `this.cacheDir = cacheDir(options) || const cache = { dir: cacheDir(options) }` 

Anyways just a nitpick 🐦 😛

src/uglify/index.js
d3viant0ne: I'd suggest that if it's renamed, we don't go with `cache` as that's used all over the place in webpack already which may / may not be confusing.

d3viant0ne:
For reference, we can publish alpha / beta builds directly from this branch. Once everyone is satisfied (enough) with it, we can push a beta for the next semver:Major & get Sean to put it out on blast via Twitter.

That has the side effect of making it easier to test the changes in the major integrations ( @angular/cli & create-react-app )

README.md
```diff
@@ -45,11 +45,12 @@ module.exports = {
 |**`test`**|`{RegExp\|Array<RegExp>}`| <code>/\.js($&#124;\?)/i</code>|Test to match files against|
 |**`include`**|`{RegExp\|Array<RegExp>}`|`undefined`|Files to `include`|
 |**`exclude`**|`{RegExp\|Array<RegExp>}`|`undefined`|Files to `exclude`|
-|**`parallel`**|`{Boolean\|Object}`|`false`|Use multi-process parallel running and file cache to improve the build speed|
+|**`cache`**|`{Boolean\|String}`|`false`|Enable file caching|
+|**`parallel`**|`{Boolean\|Number}`|`false`|Use multi-process parallel running and file cache to improve the build speed|
```
samouri: Should this line drop the part about using a file cache?

evilebottnawi: @samouri you can use own path
```
this.cacheDir = cache === true ? findCacheDir({ name: 'uglifyjs-webpack-plugin' }) : cache;
```
Just need add docs, thanks!

michael-ciniawsky:
Can we merge this PR and work on a solution for #120 ? Is it possible to rebase the current `beta` into a `develop` branch and 'revert' master to `v0.4.6` [102c18b](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/commit/102c18b95268f8dea10ba4b1fcae4d37784a95ee) since v1.0.0 will take a while (`webpack v4.0.0` ?)[Particularly talking about npm releases]. Merging into master was a mistake of mine here since `v0.4.6`  would badly need a fix for the flanky `postinstall` script and still ships with webpack atm 😞 . With `develop (@beta)`  && `master (@latest)` development doesn't get blocked by current ecosystem needs and `beta` dev can happen with less outside pressure + in the end unwanted breaking changes can be (easily) reverted before merging into master

|Branch (Github)|Tag (npm)|SemVer|
|:----|:----:|:----:|
|`develop`|`@beta`|:label: Major|
|`master`|`@latest`|:label: Minor/Patch|
|`release/vX.X.X`|`@X.X.X`| :label: Patch (only critical, non-specific)|
|`release/vX.X.X - 1`|`@X.X.X - 1`|EOL|

evilebottnawi:
@d3viant0ne @hulkish friendly ping

src/uglify/index.js
sokra: I would stay with cacache as name. name is usual equal to package name.

src/uglify/index.js
```diff
@@ -14,77 +13,69 @@ try {
...
-    if (typeof parallel === 'boolean') {
-      options = { cache: parallel, workers: parallel };
-    }
-    const { cache, workers } = options;
-    this.cache = cache === true ? findCacheDir({ name: 'uglifyjs-webpack-plugin' }) : cache;
-    this.workers = workers === true ? os.cpus().length - 1 : Math.min(Number(workers) || 0, os.cpus().length - 1);
+  constructor(options = {}) {
+    const { cache, parallel } = options;
+    this.cacheDir = cache === true ? findCacheDir({ name: 'uglifyjs-webpack-plugin' }) : cache;
+    this.maxConcurrentWorkers = parallel === true ? os.cpus().length - 1 : Math.min(Number(parallel) || 0, os.cpus().length - 1);
```
sokra: Is there a reason why only using cpu count - 1? Why not using all CPUs.

---

I would replace `Math.min(Number(parallel) || 0, os.cpus().length - 1)` with `Number(parallel)`. When the user chooses a worker count we should respect it, even if it's more than CPUs.

src/uglify/index.js
```diff
@@ -14,77 +13,69 @@ try {
...
       const enqueue = () => {
-        this.worker(task, (error, data) => {
+        this.boundWorkers(task, (error, data) => {
           const result = error ? { error } : data;
           const done = () => step(index, result);
-          if (this.cache && !result.error) {
-            put(this.cache, task.cacheKey, data, cacheIdentifier).then(done, done);
+
+          if (this.cacheDir && !result.error) {
+            cacache.put(this.cacheDir, task.cacheKey, JSON.stringify(data)).then(done, done);
```
sokra: Does it make sense to move the cache `put` operation into the worker process?

d3viant0ne:
@evilebottnawi - This in a place where you are ready for a beta build out to npm?

src/uglify/index.js
evilebottnawi: @sokra `-1` - one for master other for childs

src/uglify/index.js
evilebottnawi: @sokra no sense

evilebottnawi:
@d3viant0ne yep

src/uglify/index.js
michael-ciniawsky: Yep one core needs to be reserved for the main process running the worker farm

Genuifx:
why not merge this pr？ some of my project in window system doesn't work now. TT

d3viant0ne:
@Genuifx - It will merge when we are satisfied with it's stability. In the mean time, there are beta builds available on npm for this branch.

Genuifx:
@d3viant0ne copy that.

evilebottnawi:
@Genuifx feedback welcome :+1:

d3viant0ne:
@evilebottnawi - Updated defaults & the npm 5 / Node 4.3 issue. One the snapshots are sorted, i'd like to put out a final beta build & get this finished up this week if possible.

genintho:
>  It will merge when we are satisfied with it's stability

We have been using beta3 at work (Hipmunk) for a few days and have not experienced any issue yet. This represent around 100 build.

Issue #120, opened 2017-09-05
[Refactor] swap cacache with another cache


filipesilva:
Heya, would you be receptive to swaping the `cacache` dependency with another cache library? I can do the PR with the work for it.

I ask because I was trying to update `uglifyjs-webpack-plugin` to the latest beta in `@angular/cli`, and a couple of new licenses came up. 

Both `cacache@9.2.9` and `ssri@4.1.6` (a dependency of `cacache`) use the CC0-1.0 license (https://spdx.org/licenses/CC0-1.0.html). This license is problematic to use in Angular because it does not allow transfer of copyright and does not address some required formalities.

TheLarkInn:
This would leave me to believe this is likely incompatible with JSF license also.

evilebottnawi:
@filipesilva sounds resonably, but first we should merge https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/108, because this PR done breaking changes (separate `cache` and `parallel`), also repair logic around cache (now `cache` and `parallel` is broken and can output invalid data from cache, also `parallel` works incorrect too).

filipesilva:
@evilebottnawi SGTM, I put a https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/121 but marked it as WIP and blocked by #108.

d3viant0ne:
@TheLarkInn - Replace `likely` with `is`. I'll get it on the working list to add license checking of dependencies as a part of our validations.

mikesherov:
May I suggest using `license-to-fail` for license checking?

evilebottnawi:
@mikesherov please create issue about this in `webpack-defaults` (https://github.com/webpack-contrib/webpack-defaults)

michael-ciniawsky:
[cacache #109](https://github.com/zkat/cacache/issues/109) && [ssri #8](https://github.com/zkat/ssri/issues/8)

michael-ciniawsky:
🤞

d3viant0ne:
cacache is still overkill & quite frankly, the issue driving all of this ( performance refactoring ) is blocking the 1.5 release of the angular/cli.

We have a tool that is capable & can get the job done now, unless `flat-cache` comes with an unacceptable performance hit or is lacking in features this is moving forward as discussed in #121

TheLarkInn:
Yes let's do this

zkat:
[...]

PR #121, opened 2017-09-06
refactor: replace cacache with flat-cache

filipesilva:
See #120 for reasoning

### `Issues`

- Fixes #120

PR #145, opened 2017-10-23
fix(deps): cacache@10 with ISC licence

zkat:
Fixes: #120 

This should be able to solve the licensing issue: both `cacache` and `ssri` have been relicense to ISC (according to current npm policy), and released as a semver-major bump.

This might supersede #121

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1328,9 +1328,9 @@
       }
     },
     "cacache": {
-      "version": "9.3.0",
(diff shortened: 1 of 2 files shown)


zkat:
In re CLA: I had signed the CLA in https://raw.githubusercontent.com/webpack-contrib/uglifyjs-webpack-plugin/master/.github/PULL_REQUEST_TEMPLATE.md but that seems to be different from the one configured for this repo? (I've now signed both, so it should be fine, but maybe worth looking into)

PR #149, opened 2017-10-24
fix(compress): `options.compress` isn't assign (`webpack.optimize.UglifyJSPlugin`)

### LINKED ISSUE #146: [v0.4.6]  `options.compress` isn't assign (`webpack.optimize.UglifyJSPlugin`)
Hi, I'm using webpack 3.8.1, and after running webpack -p I have an issue when compress options doesn't work. (uglifyjs-webpack-plugin 0.4.6)
I see that compressor called multiple times and the last one he called without options (options === undefined). 
If in dist/index.js:100 I fix check with if (options.compress !== false && options.compress !== undefined) { - the last call will be with parameters so I get the expected result.

But I don't know why it called multiple times and if in webpack config will be new UglifyJSPlugin() will be empty it won't be filled with defaults.

Looks like it is a known issue, so please ask how can I help. If I can help to close this issue it would be great.

KievDevel:
Issues
Fixes #146 

Comments:
This fix probably won't fix all and this will destroy compability with previous settings. The problem is that this part of code called multiple times (in my case 3 times per file): 1st time parameters == undefined, second time they're taken from webpack.config, third time == undefined. 

Now it's called 1 time, but compress config shouldn't be empty. 
`new UglifyJSPlugin({
                compress: { }
            }),`

If it is empry it probably won't run the compressor.

Hope it will help you.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -82,7 +82,7 @@ class UglifyJsPlugin {
 						let ast = uglify.parse(input, {
 							filename: file
 						});
-						if(options.compress !== false) {
+						if(options.compress !== false && options.compress !== undefined) {
 							ast.figure_out_scope();
 							const compress = uglify.Compressor(options.compress || {
 								warnings: false


src/index.js
```diff
@@ -82,7 +82,7 @@ class UglifyJsPlugin {
 						let ast = uglify.parse(input, {
 							filename: file
 						});
-						if(options.compress !== false) {
+						if(options.compress !== false && options.compress !== undefined) {
```
michael-ciniawsky: `if (options.compress) {}` ? 

Is there a reason for checking those separately? A edge case ?

KievDevel: What I understand it shouldn't be called 3 times - only 1 time - so this case should be investigated. I suppose the real problem in this multiple calls and not in check that I proposed.

evilebottnawi:
@KievDevel friendly ping

michael-ciniawsky:
Sry but I will close and declined this, if it isn't an absolute necessary bugfix the `0.4.0` line won't recieve any futher updates and it's highly recommend to update to `v1.0.0`

PR #153, opened 2017-10-29
docs(README): update note about `webpack.optimize.UglifyJsPlugin` usage

michael-ciniawsky:


README.md
```diff
@@ -15,7 +15,7 @@
 	<p>This plugin uses <a href="https://github.com/mishoo/UglifyJS2/tree/harmony">UglifyJS v3 </a><a href="https://npmjs.com/package/uglify-es">(`uglify-es`)</a> to minify your JavaScript</p>
 </div>
 
-> ℹ️  webpack contains the same plugin under `webpack.optimize.UglifyJsPlugin`. The documentation is valid apart from the installation instructions
+> ℹ️  `webpack =< v3.0.0` currently contains [`v0.4.6`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/version-0.4) of this plugin under `webpack.optimize.UglifyJsPlugin` as an alias. For usage of the latest version (`v1.0.0`), please follow the instructions below. Aliasing `v1.0.0` as `webpack.optimize.UglifyJsPlugin` is scheduled for `webpack v4.0.0`
```
kzc: May I suggest the following wording:

> ℹ️  `webpack@3.x` makes use of [`uglifyjs-webpack-plugin@0.4.6`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/version-0.4) with `webpack.optimize.UglifyJsPlugin` as an alias. To use the latest version of this plugin, `uglifyjs-webpack-plugin@1.x`, please follow the instructions below. `webpack v4.0.0` is slated to use the `v1.x` version of this plugin with `webpack.optimize.UglifyJsPlugin` as an alias.

michael-ciniawsky: PR ? 🙃

kzc:
You may also consider adding this note as it is a frequent cause of confusion:

> ⚠️ Note: The use of `webpack -p` for `webpack@3.x` will unconditionally use `uglifyjs-webpack-plugin@0.4.6` regardless of whether `uglifyjs-webpack-plugin@1.x` was specified in `package.json` or your `webpack` config file. Instead it is recommended to use `webpack --define process.env.NODE_ENV=production` as it will work with any version of the plugin.

README.md
kzc: Sorry, just a drive by suggestion.

PR #164, opened 2017-11-15
fix(index): exclude is ignored (`options.exclude`)

rszewczyk:
A simple demo of the problem can be found [here](https://github.com/rszewczyk/uglifyjs-webpack-plugin/tree/demo-the-problem/demo).

The expected behavior is that chunks with a filename that match the [given pattern](https://github.com/rszewczyk/uglifyjs-webpack-plugin/blob/demo-the-problem/demo/webpack.config.js#L13) will be excluded from uglify. This was the behavior exhibited by __0.4.6__. The actual behavior is that all chunks are uglified.

It looks like the problem is that the [options used to filter](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/c425f3cb9612f4259af681d9c4aa1547f0d622be/src/index.js#L110) what chunks are processed, [do not include](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/c425f3cb9612f4259af681d9c4aa1547f0d622be/src/index.js#L36-L49) the exclude configuration passed to the plugin constructor.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -31,6 +31,7 @@ class UglifyJsPlugin {
       sourceMap = false,
       cache = false,
       parallel = false,
+      exclude,
     } = options;
 
     this.options = {
@@ -40,6 +41,7 @@ class UglifyJsPlugin {
       sourceMap,
       cache,
       parallel,
+      exclude,
       uglifyOptions: {
         output: {
           comments: /^\**!|@preserve|@license|@cc_on/,
--- a/test/__snapshots__/exclude-option.test.js.snap
+++ b/test/__snapshots__/exclude-option.test.js.snap
@@ -1,0 +1,62 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`errors 1`] = `Array []`;
+
+exports[`errors 2`] = `Array []`;
+
+exports[`excluded1.4d3a1b43eccbc2acc9d6.js 1`] = `
+"webpackJsonp([1],[
+/* 0 */
+/***/ (function(module, exports) {
+
+module.exports = function Bar1() {
+  const b = 2 + 2;
+  console.log(b + 1 + 2);
+};
+
+/***/ })
+],[0]);"
+`;
+
+exports[`excluded1.4d3a1b43eccbc2acc9d6.js 2`] = `
+"webpackJsonp([1],[
+/* 0 */
+/***/ (function(module, exports) {
+
+module.exports = function Bar1() {
+  const b = 2 + 2;
+  console.log(b + 1 + 2);
+};
+
+/***/ })
+],[0]);"
+`;
+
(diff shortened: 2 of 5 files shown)


everdimension:
This fix doesn't seem to work for me.

Could it be that it is because the file I wish to exclude is in `node_modules` folder?

evilebottnawi:
@everdimension can you provide configuration?

everdimension:
Well I wished I wouldn't have to make a demo repo :) But I can try later today.

But in short for now here's part of my `module` config:

```js
module: {
  rules: [
    /** ...other rules... */
    {
      test: /\.js$/,
      use: [
        {
          loader: 'babel-loader',
        },
      ],
      include: path.resolve(__dirname, 'src'),
    },
  ],
}
```

Here are uglifyjsplugin options:

```js
new UglifyJSPlugin({
  exclude: /ReactChartIQ\.translations.*\.js$/i,
  uglifyOptions: {
    ie8: false,
    warnings: false,
    sourceMap: true
  },
})
```

These work in `1.0.0-beta.1` (with according options shape)

evilebottnawi:
@everdimension strange, minimum test repo will be very helpful

rszewczyk:
I think think the issue is that the plugin [processes output files](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/index.js#L110-L113) not input files. I'm not sure about the beta, but this is the same behavior as `0.4.6`.

@everdimension I'm not sure about your use case - but we have some legacy code that can't be minified, so the way we did it (with the current version of the plugin - `0.4.6`) was to group all the code into it's own chunk by defining an entry point just for those legacy files.

evilebottnawi:
@rszewczyk yes, you are right
@everdimension move you code to entry and use `exclude`

evilebottnawi:
`test`, `include` and `exclude` options works similar `BannerPlugin` (https://github.com/webpack/webpack/blob/6a26e9ba7f7f1be7e76054f219bf1e094f2c3264/lib/BannerPlugin.js)

michael-ciniawsky:
Yep, this 'only' excludes `compilation.assets` (output), you can't exclude particular modules

everdimension:
It's not a particular module, i'm trying to exclude a filename within a module

michael-ciniawsky:
`it's not particular module <=> within a module`in the same sentence should ring a bell 😛(the import is/will be a module on it's own). At the point in the webpack lifecycle (hook) where `uglifyjs-webpack-plugin` is applied there are only the output files left without an easy way to add filters on a 'module-level' for particular modules within these assets. Why do you need this, what is the use case ?

everdimension:
> in the same sentence should ring a bell

What?

I have a particular output file name. Otherwise it wouldn't have worked before either.

PR #169, opened 2017-11-20
fix(index): ensure to nullify `inputSourceMap` if `sourcemap` is unavailable (`devtool`)

### LINKED ISSUE #111: [v1.0.0-beta.2] sourcemap not work in parallel
first set cache true, parallel false
then parallel true, the sourcemap not work

evilebottnawi:
### `Notable Changes`

- Simplify logic based on https://github.com/webpack/webpack-sources/blob/91862553dc108e897ec24ece7b5bd8aae42620e5/lib/Source.js#L24
- Add tests

### `Issues`

- Fixes #111

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -123,18 +123,16 @@ class UglifyJsPlugin {
               let input;
               let inputSourceMap;
 
-              if (this.options.sourceMap) {
-                if (asset.sourceAndMap) {
-                  const sourceAndMap = asset.sourceAndMap();
-                  inputSourceMap = sourceAndMap.map;
-                  input = sourceAndMap.source;
-                } else {
-                  inputSourceMap = asset.map();
-                  input = asset.source();
-                }
+              if (this.options.sourceMap && asset.sourceAndMap) {
+                const { source, map } = asset.sourceAndMap();
+
+                input = source;
+                inputSourceMap = map;
+
                 sourceMap = new SourceMapConsumer(inputSourceMap);
               } else {
                 input = asset.source();
+                inputSourceMap = null;
               }
 
               // Handling comment extraction
--- a/test/__snapshots__/source-map-options.test.js.snap
+++ b/test/__snapshots__/source-map-options.test.js.snap
@@ -1,0 +1,539 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`source and parallel: asset main.0c220ec66316af2c1b24.js 1`] = `"webpackJsonp([0],[function(o,n){o.exports=function(){console.log(7)}}],[0]);"`;
+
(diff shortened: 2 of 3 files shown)


test/__snapshots__/source-map-options.test.js.snap
```diff
@@ -0,0 +1,551 @@
...
+    "names": Array [
+      "module",
+      "exports",
+      "console",
+      "log",
+      "b",
+    ],
+    "sources": Array [
+      "/home/evilebottnawi/IdeaProjects/uglifyjs-webpack-plugin/test/fixtures/entry.js",
+    ],
```
d3viant0ne: This will always fail somewhere else with an absolute file path

evilebottnawi: @d3viant0ne fixed

src/index.js
```diff
@@ -123,18 +123,14 @@ class UglifyJsPlugin {
...
-                if (asset.sourceAndMap) {
-                  const sourceAndMap = asset.sourceAndMap();
-                  inputSourceMap = sourceAndMap.map;
-                  input = sourceAndMap.source;
-                } else {
-                  inputSourceMap = asset.map();
-                  input = asset.source();
-                }
+              if (this.options.sourceMap && asset.sourceAndMap) {
+                const sourceAndMap = asset.sourceAndMap();
```
```js
const { source, map } = asset.sourceAndMap();

input = source;
inputSourceMap = map;

sourceMap = new SourceMapConsumer(inputSourceMap)
```

src/index.js
```diff
@@ -123,18 +123,14 @@ class UglifyJsPlugin {
...
+                input = sourceAndMap.source;
+                inputSourceMap = sourceAndMap.map;
                 sourceMap = new SourceMapConsumer(inputSourceMap);
               } else {
                 input = asset.source();
+                inputSourceMap = null;
```
michael-ciniawsky: This is the windows related fix ?

src/index.js
michael-ciniawsky: Does `webpack >= v2.0.0` already support this convenience method ? So the `else` branch  **is** really legacy

src/index.js
evilebottnawi: @michael-ciniawsky no, when `devtool` set to null we don't have `map`

src/index.js
evilebottnawi: @michael-ciniawsky need investigate

src/index.js
michael-ciniawsky: @evilebottnawi Sure, but this is nevertheless the fix for the parallel source maps bug :) ? It doesn't seem related to `options.parallel` exclusively then, should we reframe the commit message ?

evilebottnawi: @michael-ciniawsky in this commit i just simplify logic and add tests to make sure that everything works and the problem is not in the plugin

src/index.js
michael-ciniawsky: Should be safe https://github.com/webpack/webpack/blob/webpack-2/package.json#L26

michael-ciniawsky:
@evilebottnawi Status 😛 ?  Could you address the style nit, so we can get this in or is anything blocking this PR ?

evilebottnawi:
@michael-ciniawsky done :+1:

PR #183, opened 2017-12-05
refactor: use `serialize-javascript` instead own implementation

evilebottnawi:
### `Notable Changes`

1. `serialize-javascript` more stable and support many features (see https://github.com/yahoo/serialize-javascript/blob/master/index.js#L14 https://github.com/yahoo/serialize-javascript/blob/master/index.js#L90 and etc)
2. Reducing the code, serialization out of scope plugin code and should be separate package

### `Issues`

- None


michael-ciniawsky:
Nothing against this change in particular and  if serialization is unavoidable, but I would like to traige  eventual possibilities to avoid serialization for `options` altogether if possible (maybe it isn't).

1. webpack uses an `ident`, maybe we can use a similiar approach
2. https://github.com/rvagg/node-worker-farm#exportedmethods but not 💯 if this is helpful here (didn't tested it out yet)

evilebottnawi:
@michael-ciniawsky `ident` don't contain `function`, `regex` and `date`, we can't use this as cache key

src/uglify/worker.js
```diff
@@ -1,9 +1,9 @@
 import minify from './minify';
-import { decode } from './serialization';
 
 module.exports = (options, callback) => {
   try {
-    callback(null, minify(JSON.parse(options, decode)));
+    // eslint-disable-next-line no-eval
+    callback(null, minify(eval(`(${options})`)));
```
michael-ciniawsky:
```js
options = new Function(`'use strict'\nreturn ${options}`)() // 'use strict' => this === undefined (Clean Scope)

callback(null, minify(options));
```

test/cache-options.test.js
```diff
@@ -378,11 +379,12 @@ describe('when options.cache', () => {
 
                   // Make sure that we cached files
                   expect(cacheEntriesListKeys.length).toBe(files.length);
-                  cacheEntriesListKeys.forEach((cacheJSONEntry) => {
-                    const cacheEntry = JSON.parse(cacheJSONEntry);
+                  cacheEntriesListKeys.forEach((serializedCacheEntry) => {
```
michael-ciniawsky: ` cacheEntriesListKeys` => `cacheKeys` && `serializedCacheEntry` => `cacheEntry`

test/__snapshots__/ecma.test.js.snap
```diff
@@ -106,7 +106,7 @@ exports[`ecma 6: errors 1`] = `Array []`;
 
 exports[`ecma 6: main.js 1`] = `
 "webpackJsonp([ 0 ], [ function(module, exports) {
-    class Mortgage {
+    module.exports = class {
```
michael-ciniawsky: Why did this change ? It's the same as in #190 I just can't reason about it 😛

evilebottnawi: @michael-ciniawsky oh yep :+1:

evilebottnawi:
@michael-ciniawsky done

src/uglify/worker.js
```diff
@@ -1,9 +1,14 @@
 import minify from './minify';
-import { decode } from './serialization';
 
-module.exports = (options, callback) => {
+/* eslint-disable no-new-func */
+
+module.exports = (workerOptions, callback) => {
```
michael-ciniawsky: `workerOptions` => `options` (Just reassign `options` them below please)

evilebottnawi: @michael-ciniawsky eslint throw problem about `no-param-reassign`, but it is not make sense here.

michael-ciniawsky: `eslint-disable-next-line` ? 🙃

evilebottnawi:
@michael-ciniawsky done :+1:

PR #193, opened 2017-12-17
Update: add TypeScript definitions

niedzielski:


niedzielski:
The PR template requests a CLA at https://cla.js.foundation/webpack/webpack.js.org but the bot requests https://cla.js.foundation/webpack-contrib/uglifyjs-webpack-plugin. I've now filled out both.

niedzielski:
Nuts. I missed this definition because we happen to be using the UglifyJS plugin directly per [this note](https://webpack.js.org/plugins/uglifyjs-webpack-plugin/). Unfortunately, there doesn't seem to be a workaround to use the existing definitions with the plugin directly. I tried `const UglifyJSPlugin = require("uglifyjs-webpack-plugin") as webpack.optimize.UglifyJsPlugin` but can't disambiguate the namespace from the class (I think that's [this TypeScript issue](https://github.com/Microsoft/TypeScript/issues/10544)).

PR #198, opened 2017-12-27
refactor(index): upgrade `compiler.plugin` to the new plugin system (`tapable`)

TheLarkInn:
This PR updates `uglifyjs-webpack-plugin` to use `.tap()` and `.tapAsync()` isntaed of `.plugin()` and use the new Tapable API in `webpack@webpack/webpack#next`. 

Looks like prettier did some auto updating etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "ci:lint": "npm run lint && npm run security",
     "ci:test": "npm run test -- --runInBand",
     "ci:coverage": "npm run test:coverage -- --runInBand",
-    "defaults": "webpack-defaults"
+    "defaults": "webpack-defaults",
+    "webpack-defaults": "webpack-defaults"
   },
   "dependencies": {
     "cacache": "^10.0.0",
--- a/src/index.js
+++ b/src/index.js
@@ -115,200 +115,208 @@ class UglifyJsPlugin {
   apply(compiler) {
     const requestShortener = new RequestShortener(compiler.context);
 
-    compiler.plugin('compilation', (compilation) => {
+    compiler.hooks.compilation.tap('UglifyJsWebpackPlugin', (compilation) => {
       if (this.options.sourceMap) {
-        compilation.plugin('build-module', (moduleArg) => {
-          // to get detailed location info about errors
-          moduleArg.useSourceMap = true;
-        });
+        compilation.hooks.buildModule.tap(
+          'UglifyJsWebpackPlugin',
+          (moduleArg) => {
+            // to get detailed location info about errors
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
     "ci:lint": "npm run lint && npm run security",
     "ci:test": "npm run test -- --runInBand",
     "ci:coverage": "npm run test:coverage -- --runInBand",
-    "defaults": "webpack-defaults"
+    "defaults": "webpack-defaults",
+    "webpack-defaults": "webpack-defaults"
```
michael-ciniawsky:
```diff
-   "webpack-defaults": "webpack-defaults"
```

src/index.js
```diff
@@ -115,200 +115,208 @@ class UglifyJsPlugin {
   apply(compiler) {
     const requestShortener = new RequestShortener(compiler.context);
 
-    compiler.plugin('compilation', (compilation) => {
+    compiler.hooks.compilation.tap('UglifyJsWebpackPlugin', (compilation) => {
```
michael-ciniawsky: Is `UglifyJsWebpackPlugin` accessible as a hook or is this just the plugin name for e.g interception ? `UglifyJsPlugin` ?  What's the naming convention here ? :)

d3viant0ne: @michael-ciniawsky - Check the tapable changes for 4.x. This is something you are going to see a lot of on the contrib side for Webpack 4

michael-ciniawsky: Yep, that's why I'm asking :)

TheLarkInn: This is the plugin name! :-D

package.json
TheLarkInn: Hmm how the hell did that end up in there? lol

michael-ciniawsky: It's a left over from the `webpack-defaults` `v1.6.0...2.0.0` upgrade one needs to be aware of, unrelated to this PR entirely. I just saw/catched it here :)

michael-ciniawsky:
If there aren't any objections, this should be ready to land on `next` I think (worked 💯 in #200 so far). Would be a good idea to release an `v2.0.0-alpha` in the next few days I assume, so the internet can debug (eventual) remaining bugs while `webpack` `v4.0.0` is on the finishing line :)

michael-ciniawsky:
Tests are failing because of outdated test helpers and are fixed in #200

package.json
TheLarkInn: Do I need to make a change for this? Or is PR good as is?

PR #202, opened 2018-01-03
revert(package): use exact `uglify-es` version (`dependencies`) (#199)

michael-ciniawsky:


michael-ciniawsky:
@kzc @filipesilva Does someone from the Uglify/Angular team know what exactly broke between `uglify-es v3.2.0...3.3.0` ? Is this more of an `uglify` issue in general or specific to `angular` ?

filipesilva:
@michael-ciniawsky https://github.com/mishoo/UglifyJS2/issues/2663 was the problem affecting Angular projects, and it seems JQuery as well. It was fixed via https://github.com/mishoo/UglifyJS2/commit/cb62bd98d3397d9eb3d738cc0c7f53886d3a213b and released in `3.3.3` I believe.

kzc:
I can confirm what @filipesilva said. That bug has been fixed and released.

@filipesilva Could you test Angular against the latest `uglify-es` release to verify it's working correctly?

michael-ciniawsky:
Thx @filipesilva @kzc updated the minimum to `v3.3.3`, @filipesilva Is pinning the version of `uglifyjs-webpack-plugin` in `angular-cli` instead sufficient to avoid eventual breakage through an `uglify` update in `uglifyjs-webpack-plugin` or won't that work for the `cli` ?

package.json
```diff
@@ -35,7 +38,7 @@
     "serialize-javascript": "^1.4.0",
     "schema-utils": "^0.3.0",
     "source-map": "^0.6.1",
-    "uglify-es": "3.2.2",
+    "uglify-es": "^3.3.3",
```
kzc: Any particular reason 3.3.3 was chosen?
```
     '3.2.2': '2017-12-10T18:36:46.421Z',
     '3.3.0': '2017-12-24T16:05:47.966Z',
     '3.3.1': '2017-12-25T13:11:01.187Z',
     '3.3.2': '2017-12-25T21:24:34.232Z',
     '3.3.3': '2017-12-28T21:32:12.606Z',
     '3.3.4': '2017-12-31T04:57:50.883Z' },
```
https://github.com/mishoo/UglifyJS2/commits/harmony

michael-ciniawsky: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/202#issuecomment-355142620 is the only 'reason',  I will update to `v3.3.4` (@latest) :)

michael-ciniawsky: done

filipesilva:
@michael-ciniawsky I can confirm that `3.3.3` and `3.3.4` do not show the problem affecting Angular applications.

Also I don't think this is intended, but you're also reverting the `1.1.5` release in this PR.

michael-ciniawsky:
> Also I don't think this is intended, but you're also reverting the 1.1.5 release in this PR.

@filipesilva 😱 Thx 👍

PR #207, opened 2018-01-08
docs(README): `mangle.props` instead of `mangle.properties`

davilima6:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. See [standa
 
 ### Bug Fixes
 
+* Fix typo: mangle properties key should be props
```
michael-ciniawsky: Please revert this change as the `CHANGELOG` is generated and the PR isn't part of this particular release :)

davilima6: Thanks for the advice but I think this PR is not correct (it changed from 0.4.6 to 1.x)

michael-ciniawsky: The `CHANGELOG` is still auto generated from commit messages and your PR isn't part of `v1.1.6` release, so this needs to be reverted please :)

davilima6: done

kzc:
This PR is incorrect. Please revert.

The uglify `minify()` `mangle` sub-option is `properties`, not `props`.

See: https://github.com/mishoo/UglifyJS2/tree/harmony#minify-options-structure

michael-ciniawsky:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/commit/44b77e3b2bec5d63baf340c382d6564123a4d108

PR #227, opened 2018-02-07
fix: add missing `keep_classnames`, `keep_fnames` && `safari10` options (`uglifyOptions`)

vepanimas:
There are omited top-level options `keep_classnames`, `keep_fnames` and `safari10` for uglify-js. I believe we should be able to use all options supported by underlying minification package.


evilebottnawi:
@vepanimas Thanks for PR. Close if favor https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/229

michael-ciniawsky:
Released in [`v1.2.0`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases/tag/v1.2.0) 🎉  via #229  @vepanimas Thx for bringing this up 👍

PR #233, opened 2018-02-15
feat(minify): extract `LICENSE && Copyright` comments by default (`options.extractComments`)

evan-scott-zocdoc:
This is the regex we've been using in our builds and it's been successfully grabbing more of the license headers.

### DIFF
--- a/src/uglify/minify.js
+++ b/src/uglify/minify.js
@@ -40,7 +40,7 @@ const buildComments = (options, uglifyOptions, extractedComments) => {
   // /^\**!|@preserve|@license|@cc_on/
   if (typeof options.extractComments === 'boolean') {
     condition.preserve = commentsOpts;
-    condition.extract = /Copyright|LICENSE|@preserve|@license|@cc_on/ig;
+    condition.extract = /^\**!Copyright|LICENSE|@preserve|@license|@cc_on/ig;
   } else if (
     typeof options.extractComments === 'string' ||
     options.extractComments instanceof RegExp
--- a/src/uglify/minify.js
+++ b/src/uglify/minify.js
@@ -40,7 +40,7 @@ const buildComments = (options, uglifyOptions, extractedComments) => {
   // /^\**!|@preserve|@license|@cc_on/
   if (typeof options.extractComments === 'boolean') {
     condition.preserve = commentsOpts;
-    condition.extract = /^\**!Copyright|LICENSE|@preserve|@license|@cc_on/ig;
+    condition.extract = /^\**!|Copyright|LICENSE|@preserve|@license|@cc_on/ig;
   } else if (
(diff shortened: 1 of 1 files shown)


src/uglify/minify.js
```diff
@@ -40,7 +40,7 @@ const buildComments = (options, uglifyOptions, extractedComments) => {
   // /^\**!|@preserve|@license|@cc_on/
   if (typeof options.extractComments === 'boolean') {
     condition.preserve = commentsOpts;
-    condition.extract = /^\**!|@preserve|@license|@cc_on/;
+    condition.extract = /Copyright|LICENSE|@preserve|@license|@cc_on/ig;
```
evilebottnawi: Don't remove `\**!`

evilebottnawi:
@evan-scott-zocdoc Thanks for PR

src/uglify/minify.js
evan-scott-zocdoc: The regex won't work without it, since not everyone uses the `/**` type of comments.

evilebottnawi: @evan-scott-zocdoc we use `|` (`or`) for comments types

evan-scott-zocdoc: OHH, sorry I see. Will fix it.

src/uglify/minify.js
```diff
@@ -40,7 +40,7 @@ const buildComments = (options, uglifyOptions, extractedComments) => {
   // /^\**!|@preserve|@license|@cc_on/
   if (typeof options.extractComments === 'boolean') {
     condition.preserve = commentsOpts;
-    condition.extract = /^\**!|@preserve|@license|@cc_on/;
+    condition.extract = /^\**!|Copyright|LICENSE|@preserve|@license|@cc_on/ig;
```
michael-ciniawsky: Are these `uglify-es` defaults aswell ?

evan-scott-zocdoc: Not sure, this is the config we've been using at Zocdoc that extracts more of the relevant comments.

evilebottnawi: @evan-scott-zocdoc maybe also do PR in uglify-es?

evan-scott-zocdoc: I could, but you're already defining your own override here so me changing it in uglify-es wouldn't fix anything?

evilebottnawi: @evan-scott-zocdoc First in uglify, second we merge here

michael-ciniawsky: If I remember right for `options.extractComments` which is a feature of `uglifyjs-webpack-plugin`, we took those defaults from `uglify-es` (default preserved comments). It's better to keep `uglifyjs-webpack-plugin` as near to `uglify-es` defaults as possible even for a small change like this.

PR #243, opened 2018-02-26
perf(index): use `md4` instead `md5` for content hasing

evilebottnawi:
Ref: https://github.com/webpack/webpack/pull/6549


michael-ciniawsky:

PR #244, opened 2018-02-26
fix(uglify/worker): better `{Error}` handling

evilebottnawi:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -183,6 +183,7 @@ class UglifyJsPlugin {
 
           // Handling results
           // Error case: add errors, and go to next file
+          // Using when we don't use parallel
           if (error) {
             compilation.errors.push(
               UglifyJsPlugin.buildError(
--- a/src/uglify/worker.js
+++ b/src/uglify/worker.js
@@ -7,7 +7,9 @@ module.exports = (options, callback) => {
     // eslint-disable-next-line no-new-func, no-param-reassign
     options = new Function(`'use strict'\nreturn ${options}`)();
 
-    callback(null, minify(options));
+    const result = minify(options);
+
+    callback(result.error, result);
   } catch (errors) {
     callback(errors);
   }


michael-ciniawsky:
Could you please show a screenshot of the output ? Nice if it already works with this elegant change :)

michael-ciniawsky:
Also the `{Error}` shape/type of the `err` before pushing to `compilation.errors` needs to be checked :)

evilebottnawi:
@michael-ciniawsky what do you mean? :smile:

michael-ciniawsky:
A screenshot/logs of the {Error} and what does the worker return as `err` `{String|Object}` `{Object} === instanceof Error` (when it is passed to `compilation.errors.push(err)`) ? 🙃

michael-ciniawsky:
Could we clearify if this definitely fixes the `{Error}` handling and get this PR and #243 out asap ? :)

evilebottnawi:
@michael-ciniawsky :+1:

evilebottnawi:
@michael-ciniawsky also remove error handle, because we already push error

evilebottnawi:
@michael-ciniawsky thos solution don't works because we can't build `UglifyJsPlugin.buildError` we should serialize error for right account :disappointed:

michael-ciniawsky:
Can you explain this a bit more detailed please :) ? The `buildUglifyError` could maybe made it's own`{Error}` Class instead, which handles the serialized error `{String|Object}`

**Uglify[Worker]Error.js**
```js
class Uglify[Worker]Error extends Error {
   constructor (serialized) {
      const err = JSON.parse(serialized)
       
      this.name = err.name
      this.message = `\n\n(${err.line}:${err.column} ${err.message}...\n`
      
      Error.captureStackTrace(this, this.constructor)
   }
}

module.exports = Uglify[Worker]Error
```

d3viant0ne:
Please rebase with current master

evilebottnawi:
@michael-ciniawsky apparently something was wrongly installed or configured, the problem is no longer there

PR #200, opened 2017-12-27
refactor(test): standardize

michael-ciniawsky:
### `Notable Changes`

- Removes the `plugin` test helpers
- Cleanup and simplifies various tests

### `Issues` 

- None

### `Dependencies`

- #198 

> ⚠️ This is needed for the tests to work with the new plugin system (`webpack >= v4.0.0`)

PR #250, opened 2018-03-06
fix(index): don't remove legal comments by default (`options.extractComments`)

### LINKED ISSUE #222: Legal, License, Copyright comments should not be removed by default
**Steps to reproduce:**

Version in `package.json`:
```json
    "grunt-webpack": "^3.0.2",
    "uglifyjs-webpack-plugin": "^1.1.8",
    "webpack": "^3.10.0",
    "webpack-bundle-analyzer": "^2.9.2"
```

My `webpack.config.js`:
```js
const path = require('path');
const webpack = require('webpack');
const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
  .BundleAnalyzerPlugin;

const config = {
  // entry: ["babel-polyfill", "./app/js"]
  entry: {
    // Further entry points can be added here
    // NB: If you add further ones, then you probably also want to move
    // dependencies like jQuery into their own bundle
    // see https://webpack.js.org/guides/code-splitting/#prevent-duplication

    // the babel-polyfill and fetch polyfill needs to be loaded only once,
    // hence only added to app.js which is executed on every page
    app: ['babel-polyfill', 'whatwg-fetch', './src/components/app.js'],
    resizer: ['./src/prototype-assets/viewport-resizer/main.js'],
    fractalmenuenhancer: [
      './src/prototype-assets/fractal-menu-enhancer/main.js'
    ]
  },
  module: {
    rules: [
      {
        test: /\.js$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader'
        }
      }
    ]
  },
  resolve: {
    extensions: ['*', '.js', '.jsx']
  },
  output: {
    filename: '[name].bundle.js',
    path: path.resolve(__dirname, 'public/js/')
  },
  plugins: [
    new webpack.EnvironmentPlugin({
      // pass 'development' unless process.env.NODE_ENV is defined to further tasks/loaders/babel
      NODE_ENV: 'development'
    })
  ]
};

// Extend config to use best, small build for prod and fast, big for dev
if (process.env.NODE_ENV === 'production') {
  config.devtool = 'source-map';
  config.plugins.push(
    new UglifyJSPlugin({
      sourceMap: true
    })
  );
  config.plugins.push(
    new BundleAnalyzerPlugin({
      analyzerMode: 'static',
      reportFilename: '../webpack-bundle-size-analyser.html',
      openAnalyzer: false
    })
  );
} else {
  config.devtool = 'eval-source-map';
}

module.exports = config;
```

Now in the `./src/components/app.js` there are several 'vendor' libraries imported like: `import $ from 'jquery';`

**Actual results:**
When running `npm run build` which creates a uglified minimal bundle, then *all* comments are stripped.

For instance, `jquery` has this header which is not present any more in the uglified bundle:
```
/*!
 * jQuery JavaScript Library v3.3.1
 * https://jquery.com/
 *
 * Includes Sizzle.js
 * https://sizzlejs.com/
 *
 * Copyright JS Foundation and other contributors
 * Released under the MIT license
 * https://jquery.org/license
 *
 * Date: 2018-01-20T17:24Z
 */
``` 

**Expected results:**
Legal copyright comments should not be stripped by default. The comment at https://github.com/webpack/webpack/issues/324#issuecomment-157855568 points out, that

> it's intended that webpack preserves some comments. This is for legal reasons. By default comments with @license, @preserve or starting with /*! are preserved

At the moment this is not the case: By default *all* comments are stripped at the moment.

evilebottnawi:
### `Issues`

- Fixes #222

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ class UglifyJsPlugin {
       exclude,
       uglifyOptions: {
         output: {
-          comments: false,
+          comments: extractComments ? false : /^\**!|@preserve|@license|@cc_on/,
         },
         ...uglifyOptions,
       },
--- a/test/extract-comments-options.test.js
+++ b/test/extract-comments-options.test.js
@@ -2,10 +2,7 @@ import UglifyJsPlugin from '../src/index';
 import { PluginEnvironment } from './helpers';
 
 describe('when options.extractComments', () => {
-  let compilation;
-  let compilationEventBinding;
-
-  beforeEach(() => {
+  it('normalizes extractConmments', () => {
     const pluginEnvironment = new PluginEnvironment();
     const compilerEnv = pluginEnvironment.getEnvironmentStub();
     compilerEnv.context = '';
@@ -26,7 +23,7 @@ describe('when options.extractComments', () => {
     const [eventBinding] = pluginEnvironment.getEventBindings();
     const chunkPluginEnvironment = new PluginEnvironment();
 
-    compilation = chunkPluginEnvironment.getEnvironmentStub();
+    const compilation = chunkPluginEnvironment.getEnvironmentStub();
     compilation.assets = {
       'test.js': {
         source: () => 'var foo = 1;',
@@ -47,10 +44,8 @@ describe('when options.extractComments', () => {
     compilation.errors = [];
 
     eventBinding.handler(compilation);
-    [compilationEventBinding] = chunkPluginEnvironment.getEventBindings();
-  });
+    const [compilationEventBinding] = chunkPluginEnvironment.getEventBindings();
 
-  it('normalizes extractConmments', () => {
     compilationEventBinding.handler([{
       files: ['test.js'],
     }], () => {
@@ -59,6 +54,49 @@ describe('when options.extractComments', () => {
   });
 
   it('outputs warnings for unreachable code', () => {
+    const pluginEnvironment = new PluginEnvironment();
+    const compilerEnv = pluginEnvironment.getEnvironmentStub();
+    compilerEnv.context = '';
+
+    const plugin = new UglifyJsPlugin({
+      uglifyOptions: {
+        warnings: true,
+        mangle: {
+          properties: {
+            builtins: true,
+          },
+        },
+      },
+      extractComments: true,
+    });
+    plugin.apply(compilerEnv);
+
+    const [eventBinding] = pluginEnvironment.getEventBindings();
+    const chunkPluginEnvironment = new PluginEnvironment();
+
+    const compilation = chunkPluginEnvironment.getEnvironmentStub();
+    compilation.assets = {
+      'test.js': {
+        source: () => 'var foo = 1;',
+      },
+      'test1.js': {
+        source: () => 'function foo(x) { if (x) { return bar(); not_called1(); } }',
+        map: () => {
+          return {
+            version: 3,
+            sources: ['test1.js'],
+            names: ['foo', 'x', 'bar', 'not_called1'],
+            mappings: 'AAAA,QAASA,KAAIC,GACT,GAAIA,EAAG,CACH,MAAOC,MACPC',
+          };
+        },
+      },
+    };
+    compilation.warnings = [];
+    compilation.errors = [];
+
+    eventBinding.handler(compilation);
+    const [compilationEventBinding] = chunkPluginEnvironment.getEventBindings();
+
     compilationEventBinding.handler([{
       files: ['test.js', 'test1.js'],
     }], () => {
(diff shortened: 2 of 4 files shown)


test/extract-comments-options.test.js
```diff
@@ -85,7 +122,7 @@ describe('when options.extractComments', () => {
         source: () => '// Comment\nvar foo = 1;',
       },
       'test1.js': {
-        source: () => '/* Comment */\nvar foo = 1;',
+        source: () => '/*! Legal Comment */\nvar foo = 1;',
```
evilebottnawi: Also fix basic tests, because `/* Comment */` is not legal comment and `test1.js.LICENSE` empty, but we expect below what `test1.js.LICENSE` is not empty

evilebottnawi:
Also need refactor tests, but we can do this before next minor/major release, i will send PR in future

src/index.js
```diff
@@ -22,7 +22,7 @@ class UglifyJsPlugin {
       uglifyOptions = {},
       test = /\.js$/i,
       warningsFilter = () => true,
-      extractComments = false,
+      extractComments = true,
```
michael-ciniawsky: This would produce a `name.js.LICENSE` file by default or just preserves the comments within the file ? Is it really necessary to produce this file by default instead of opt-in if needed ? 

`semver`-ness is also debatable here...

michael-ciniawsky:
> Also need refactor tests, but we can do this before next minor/major release, i will send PR in future

#200 (needs a rebase)

src/index.js
evilebottnawi: @michael-ciniawsky yep, it is output `name.js.LICENSE` file, same as in `@0.4.6` version

michael-ciniawsky: Was outputting the `file.js.LICENSE` the default in `v0.4.6` ?  I can't remember... 😄 and if so do we really want it to be the default ?

evilebottnawi: @michael-ciniawsky ok let's test this :smile: i will put screen shot :+1:

evilebottnawi: @michael-ciniawsky i was right :smile: 
screenshot from 2018-03-07 22-08-27

evilebottnawi:
```js
new UglifyJSPlugin({
    extractComments: true
})
```

michael-ciniawsky: kk, maybe asking folks (e.g slack) about their preference (on by default Y/n ?) would be a good idea, if there is no response or people want then let's move ahead

evilebottnawi: @michael-ciniawsky Good idea

evilebottnawi: @michael-ciniawsky from slack
```
sokra [4:27 PM]
I would keep comments in code by default. LICENSE file by default could be confusing for users and the benefit is small...
```

evilebottnawi: Let's do it?

julmot: Ping @evilebottnawi @michael-ciniawsky

evilebottnawi: @julmot WIP :+1:

evilebottnawi:
@michael-ciniawsky friendly ping. Also:
- Refactor tests for readability.
- Add tests when `extractComments` is not specify

src/index.js
```diff
@@ -42,7 +42,7 @@ class UglifyJsPlugin {
       exclude,
       uglifyOptions: {
         output: {
-          comments: false,
+          comments: extractComments ? false : /^\**!|@preserve|@license|@cc_on/,
```
michael-ciniawsky: Shouldn't this be just the `{RegExp}` to always preserve this comments ? Why is the `extractComments` check needed here ?

evilebottnawi: @michael-ciniawsky When we `extractComments` comments we should remove their from source code. If you pass `extractComments: true` and `comments: /^\**!|@preserve|@license|@cc_on/` output will be contain legal comments in source file and generate `name.js.LICENSE` file

michael-ciniawsky: Isn't  `extractComments` not only responsible for **only** extracting what's preseverd via `uglifyOptions.output.comments` ?

```js
// Preserve comments based on options, defaults to /^\**!|@preserve|@license|@cc_on/
if (uglifyOptions.output.comments) {
   // Extract all preserved comments
   if (options.extractComments) {
       // [name].js.LICENSE
       extractComments()
   }
   // Leave comments in the source 
}
```

Why is are these options 'decoupled' ? 🙃 

Anyways if that's more complicated atm and this PR now leaves the 'default'  preserved comments in the source aswell as `options.extractComments` also preserves the same default comments (with it's own logic) when extracting them into a separate file, then let's do it :)

evilebottnawi: @michael-ciniawsky `extractComments` does not responsible uglify comments, they are separatly. If think about your logic. But sometimes people leave comment in source code and generate license file to for allow difference type of distribution.

psulek:
@evilebottnawi In which version of `uglifyjs-webpack-plugin` will this be published? In current latest version `1.2.3` it does not work.

michael-ciniawsky:
@evilebottnawi There is nothing left to do here for the moment right ? :) In case there is please chime in, otherwise I will land and release this today...

evilebottnawi:
@michael-ciniawsky let's release :+1:

michael-ciniawsky:
Released in [`v1.2.4`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases/tag/v1.2.4) 🎉

psulek:
@michael-ciniawsky at npm there is still `1.2.3` version. Am i too fast ?

michael-ciniawsky:
https://github.com/npm/npm/issues/20077 😞

PR #251, opened 2018-03-06
fix(index): relax default in asset names `{RegExp}` (`options.test`)

### LINKED ISSUE #220: uglifyjs-webpack-plugin cannot uglify any file
this is my code:
```
new UglifyJsPlugin({
    uglifyOptions: {
         ie8: false,
	 ecma: 8,
	 //parse: {...options},
	 mangle: {
		 properties: {
		          // mangle property options
		  }
	  },
	 output: {
		   comments: false,
		   beautify: false
	  },
	 compress: {
	  },
	 warnings: false
    }
 })
```
 I used es7, es6, if i use webpack.optimize.UglifyJsPlugin , it can work normally.

```
"uglifyjs-webpack-plugin": "^1.1.8",   
 "webpack": "^3.10.0"
```

evilebottnawi:
### `Issues`

- Fixes #220

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ class UglifyJsPlugin {
 
     const {
       uglifyOptions = {},
-      test = /\.js$/i,
+      test = /\.(js?)(\?.+)?$/i,
       warningsFilter = () => true,
       extractComments = false,
       sourceMap = false,
--- a/test/__snapshots__/test.test.js.snap
+++ b/test/__snapshots__/test.test.js.snap
@@ -1,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`empty option: 0.0.js?ver=bd0395a3b5883cc6e11c 1`] = `"webpackJsonp([0],{5:function(e,t,c){\\"use strict\\";Object.defineProperty(t,\\"__esModule\\",{value:!0}),t.default=\\"async-dep\\"}});"`;
+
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -20,7 +20,7 @@ class UglifyJsPlugin {
 
     const {
       uglifyOptions = {},
-      test = /\.js$/i,
+      test = /.((js)?)(\?[a-z0-9]+)?$/i,
```
michael-ciniawsky: The `.` needs to be escaped (?)

```diff
- /.js
+ /\.js...
```

Why `((js)?)` ('nested' capturing groups) ?

evilebottnawi:
@michael-ciniawsky done :+1:

michael-ciniawsky:

jazdw:
@evilebottnawi was it intentional to match the .j extension? Also an empty query should probably be allowed. Maybe `\.js(\?.*)?$`?

michael-ciniawsky:
> was it intentional to match the .j extension?

The common **default** at this stage of the build is a JS asset ending in `.js` (`output.filename/output.chunkFilename`), it's possible to override this via `options.test` in case it's needed (but **optional**)

> Also an empty query should probably be allowed. Maybe \.js(\?.*)?$?

What's the 'use case' for an `empty` query? Why/When would someone do this ? But the `{RegExp}` could eventually be further relaxed/changed if there is a case made :)

jazdw:
I'm not following you, maybe I didn't make myself clear. The Regex in the commit will match files with a .js and a .j extension which I don't think is desirable or the intention.

There's not a use case per say, just in case someone uses a variable substitution that evaluates to an empty string.

jazdw:
@evilebottnawi @michael-ciniawsky any comment on the Regex?

evilebottnawi:
@jazdw oh, yes, you are right we should fix it :+1:

PR #275, opened 2018-04-09
refactor: update `jest` and tests

evilebottnawi:


evilebottnawi:
@d3viant0ne looks like update `jest` is not solve problem :confused: investigate deeply

.circleci/config.yml
```diff
@@ -1,19 +1,20 @@
...
     - checkout
-    - setup_remote_docker
     - restore_cache:
         key: dependency-cache-{{ checksum "package-lock.json" }}
     - run:
         name: NPM Rebuild
         command: npm install
+    - run:
+        name: Build project.
+        command: npm run build
```
evilebottnawi: Require for https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/uglify/index.js#L12

PR #276, opened 2018-04-10
perf: reduce memory usage

evilebottnawi:
We create source map on each `file` and it is increase memory usage very very very lagre, but using `sourceMap` only for error. Let's create `sourceMap` only when erros/warnings were throws.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,11 +48,18 @@ class UglifyJsPlugin {
         ...uglifyOptions,
       },
     };
+    this.sourceMapsCache = new WeakMap();
   }
 
-  static buildError(err, file, sourceMap, requestShortener) {
+  buildError(err, file, inputSourceMap, requestShortener) {
     // Handling error which should have line, col, filename and message
     if (err.line) {
+      const sourceMapCacheKey = { file };
+      let sourceMap = this.sourceMapsCache.get(sourceMapCacheKey);
+      if (!sourceMap) {
+        sourceMap = new SourceMapConsumer(inputSourceMap);
+        this.sourceMapsCache.set(sourceMapCacheKey, sourceMap);
+      }
       const original = sourceMap && sourceMap.originalPositionFor({
         line: err.line,
         column: err.col,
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@TheLarkInn no issue, it is own metrics in our private projects. We speedup plugins (because now we don't spend time on parse source maps everytime, because we need in they only when error occurs), also don't use memory for this.

src/index.js
```diff
@@ -66,21 +73,29 @@ class UglifyJsPlugin {
...
-      const original = sourceMap.originalPositionFor({
+      const original = sourceMap && sourceMap.originalPositionFor({
         line,
         column,
       });
 
       if (original && original.source && original.source !== file && warningsFilter(original.source)) {
         accWarnings.push(`${warning.replace(warningRegex, '')}[${requestShortener.shorten(original.source)}:${original.line},${original.column}]`);
+      } else {
+        accWarnings.push(`${warning.replace(warningRegex, '')}`);
```
evilebottnawi: Fix, when we don't have source map warnings not emitting :confused:

test/__snapshots__/all-options.test.js.snap
```diff
@@ -57,7 +57,10 @@ exports[`when applied with all options matches snapshot: manifest.d6857f782c13a9
 
 exports[`when applied with all options matches snapshot: warnings 1`] = `
 Array [
-  "Error: main.0c220ec66316af2c1b24.js from UglifyJs
-Dropping unused variable a [./test/fixtures/entry.js:4,0]",
+  Array [
+    "Error: main.27965655b499b10158f3.js from UglifyJs",
+    "Dropping unused variable a [./test/fixtures/entry.js:4,0]",
+    "Dropping unused variable c [./test/fixtures/entry.js:5,0]",
```
evilebottnawi: Fix error helper in tests. Now we can see all warnings in tests.

evilebottnawi:
@d3viant0ne need patch release :+1:

BoldBigflank:
I'm trying out 1.2.5 and still seeing a hang at 92% (I thought it was working but it was just failing at a missing dependency at first)

PR #280, opened 2018-04-26
Don't use a worker farm unless more than one process is required

hedgepigdaniel:
This should save some memory due to sharing common data within one process instead of spawning a second one.

### DIFF
--- a/src/uglify/index.js
+++ b/src/uglify/index.js
@@ -25,7 +25,7 @@ export default class {
       return;
     }
 
-    if (this.maxConcurrentWorkers > 0) {
+    if (this.maxConcurrentWorkers > 1) {
       const workerOptions = process.platform === 'win32' ? { maxConcurrentWorkers: this.maxConcurrentWorkers, maxConcurrentCallsPerWorker: 1 } : { maxConcurrentWorkers: this.maxConcurrentWorkers };
       this.workers = workerFarm(workerOptions, workerFile);
       this.boundWorkers = (options, cb) => this.workers(serialize(options), cb);


hedgepigdaniel:
The difference is for a two core system - since the default number of workers is `numcpus - 1`. In our case we have `parallel: true` in the config, and this causes our two core build server to run two processes even though only one of them is actually doing the minifying.

Tbh, I haven't tested that is uses less memory but I assume that there is at least some memory that is shared in one process that needs to be duplicated in a second one. I tested that it works though :-1:

evilebottnawi:
@hedgepigdaniel Looks you are right today, i am testing memory usage and i think we can merge this :+1:

evilebottnawi:
@michael-ciniawsky `os.cpus().length`:
1. When we have one core (`os.cpus().length === 1`) - we create worker, it is wrong
2. When we have two core (`os.cpus().length === 2`) - we create two worker, it is wrong (one thread should be left for webpack and don't create new thread because we have 1 thread, it is does make sense, only decrease memory/cpu for creating one thread)
3. When we have fore core (`os.cpus().length === 4`) - we create 4 worker, it is wrong (one thread should be left for webpack and 3 for uglify)

michael-ciniawsky:
The `worker-farm` docs don't state that there must be one core 'freed' for the 'main' process, but maybe you're right, as mentioned worth a **try** :).

evilebottnawi:
@michael-ciniawsky it is good practice, If at least one thread completed it is require `webpack` main thread to create new thread for other file (otherwise we will lose time).

evilebottnawi:
@michael-ciniawsky Also using `os.cpus().length` create new problem for one core system (other used in CI system). We create one extra thread on 1 core system and spend extra resources (memory and cpu). It does not make sense to create a new thread on one core system

michael-ciniawsky:
@evilebottnawi We don't certainly know for sure if this would be the case or not, I'm not disagreeing but by at least trying we might get proof of the circumstances here

evilebottnawi:
@michael-ciniawsky why we should create new thread on one core system? It is does not have any sense

evilebottnawi:
Each creating new thread using more memory and cpu

evilebottnawi:
@michael-ciniawsky You can justify in detail what you do not agree with PR, or approve PR

evilebottnawi:
@michael-ciniawsky What exactly do you disagree? `os.cpus().length - 1` is good practice for all multi threading platform, do you have knowledge about implement multi threading or have experience, if not it does not make sense to participate in the discussion and speedy solution of the problem. Each prgorams running in multithreading mode has `master` process and we should always leave one thread for master (creating new threads, send signal for other process, regular work)

evilebottnawi:
In short: 
- you have 8 files to uglify
- you system support 4 threads.
- plugin create 4 thread for uglification
- 1,2,3 thread completed uglification, if we don't leave one thread for master, no new threads for other 4 files, because master can't create new threads. We will have to wait until we finish 4 thread complete you works, because create new thread places in master code https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/uglify/index.js#L53

d3viant0ne:
Agree with @evilebottnawi here. If you don't have one master thread, you end up context switching all over the place and take a huge performance hit and may as well not thread it at all.

Reserving one thread to orchestrate execution has been a generally accepted practice pretty much since we started leveraging the power of multi-core processors.

evilebottnawi:
@d3viant0ne can we merge this? i don't have administration right here

PR #284, opened 2018-05-02
fix: improve `isSouceMap` check

### LINKED ISSUE #283: Unnecessary warning on "invalid source map" for JSON files
Hi everyone :)

When `import()`ing a json file, the generated source map is valid but "empty", which causes https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/268/files#diff-8968c62593320351ced6a0ef776ef599R8 to fail and emit a warning.

See https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/268

**Example:**

`data.json`
```json
{
	"hello": "world"
}
```

generates the following source map:

```json
{
 "version":1,
 "sources":[],
 "names":[],
 "mappings":"",
 "file":"data.8efcc96623969177ff47.js",
 "sourceRoot":""
}
```

Unfortunately this causes `isSourceMap()` from https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/268/files#diff-8968c62593320351ced6a0ef776ef599 to not recognize it as a valid source map. (`mappings` check fails).

Even when using `SourceMapDevToolPlugin()` https://webpack.js.org/plugins/source-map-dev-tool-plugin/ and explicitely excluding source maps for such files the warning will still be added. This should definitely not happen, I guess? Of course there is "no valid source map", because there is no source map at all.

Ultimately it adds the following warning to the webpack compiler output:
> data.8efcc96623969177ff47.js contain invalid source map

Sidenote: There is a typo, it should be "contains".

Side-Sidenote: This "warning" comes without any "error stack" and is really, really hard to debug. Also searching on google / stackoverflow etc. is extremely hard because it's so generic and there is no way of knowing who warned me. Is it webpack? Babel? Typescript?

I only found the cause by change after intense searching and almost losing my mind. 🤪
Would love to see some improved message here too :)

Cheers and thank you,
Stefan

evilebottnawi:
fixes #283

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,7 +149,7 @@ class UglifyJsPlugin {
               } else {
                 inputSourceMap = map;
                 compilation.warnings.push(
-                  new Error(`${file} contain invalid source map`),
+                  new Error(`${file} contains invalid source map`),
                 );
               }
             } else {
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -4,8 +4,8 @@ function isSourceMap(input) {
   return Boolean(input &&
     input.version &&
     input.sources &&
-    input.names &&
-    input.mappings);
+    Array.isArray(input.sources) &&
+    typeof input.mappings === 'string');
 }
 
 export default {
--- a/test/__snapshots__/source-map-options.test.js.snap
+++ b/test/__snapshots__/source-map-options.test.js.snap
@@ -4,7 +4,7 @@ exports[`when options.sourceMap true and options.parallel true compilation handl
 
 exports[`when options.sourceMap true and options.parallel true compilation handler when called optimize-chunk-assets handler only calls callback once: warnings 1`] = `
 Array [
-  [Error: test4.js contain invalid source map],
+  [Error: test4.js contains invalid source map],
 ]
 `;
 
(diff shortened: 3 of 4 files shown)


nehaabrol87:
@sokra  With what version of webpack 4 is this plugin available ? When mode = 'production' , webpack 4 automatically includes UglifyJsPlugin . Ever since I upgraded my preact project to use webpack 4 , my development build was working fine but some react beatutiful dnd modules were being rendered as undefined only in production.

After going down a lot of rabbit holes , I figured setting mode = development on production webpack config works but of course I still do not understand if the real problem is webpack 4 or preact-compat not working well with any of the automatically included plugins in webpack 4 -

**Plugins automatically available with webpack 4**

Provides process.env.NODE_ENV with value production. Enables FlagDependencyUsagePlugin, FlagIncludedChunksPlugin, ModuleConcatenationPlugin, NoEmitOnErrorsPlugin, OccurrenceOrderPlugin, SideEffectsFlagPlugin and UglifyJsPlugin

**Setting that works in my project and does not render undefined for react-beautiful-dnd modules**
 
mode = 'development'
**and adding**
  optimization: {
    namedModules: true, // NamedModulesPlugin()
    splitChunks: { // CommonsChunkPlugin()
        name: 'vendor',
        minChunks: 2
    },
    noEmitOnErrors: true, // NoEmitOnErrorsPlugin
    concatenateModules: true
},

**NOTE**: Long story short , I still do not understand  if the real problem is webpack 4 or preact-compat not working well with any of the automatically included plugins in webpack 4.
I am going too notify preact-compat folks as well.

**Versions**
webpack@4.8.1 
preact-compat@3.18.0

Issue #286, opened 2018-05-04
Slow builds when there are many warnings/errors


lojjic:
This issue is to continue discussion from: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/commit/0b11134fea8e92e92ec1b4b9bfe3cdcc9d8e640b#r28859767

Version 1.2.5 increased our build time by 4x, and I tracked it down to the above commit which implemented a sourcemap cache in a way that cache hits are impossible. Fixing the cache misses brings our build time back to normal, however there is some question whether this fix should be done with a `WeakMap` to allow GC or just use a simple `Map`.

lojjic:
Another possible option (if I'm understanding the code flow correctly): remove the cache object altogether, go back to creating the `sourceMap = new SourceMapConsumer(...)` outside the loop, but make it so that it's created lazily only after we know there are errors or warnings in the result. Seems like this would meet the goals as I understand them: it's only created once per file, it's only created if there are actually errors or warnings, and it can be GC'd as soon as its variable pointer is out of scope.

evilebottnawi:
@lojjic PR welcome, let's discussion on code

PR #287, opened 2018-05-04
perf(src/index): regression caused by parsing source maps individually

lojjic:
See issue #286 

This approach reverts to not using a separate cache object, but being careful about lazily instantiating the `SourceMapConsumer` only when we know it is needed.

In my project's build, this took the number of instantiated `SourceMapConsumer` objects from 1418 down to 21.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,18 +48,11 @@ class UglifyJsPlugin {
         ...uglifyOptions,
       },
     };
-    this.sourceMapsCache = new WeakMap();
   }
 
-  buildError(err, file, inputSourceMap, requestShortener) {
+  static buildError(err, file, sourceMap, requestShortener) {
     // Handling error which should have line, col, filename and message
     if (err.line) {
-      const sourceMapCacheKey = { file };
-      let sourceMap = this.sourceMapsCache.get(sourceMapCacheKey);
-      if (!sourceMap) {
-        sourceMap = new SourceMapConsumer(inputSourceMap);
-        this.sourceMapsCache.set(sourceMapCacheKey, sourceMap);
-      }
       const original = sourceMap && sourceMap.originalPositionFor({
         line: err.line,
         column: err.col,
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@lojjic local tests, in near future i will do pr with benchmark npm command. We need source map only when we have warning/errors. Right now you solution always try to parse source map even i don't have warning and error. I.E. increase memory usage when i don't need this.

evilebottnawi:
Just replace `WeakMap` on `Map` it is allow to find golden way for all (who has errors/warnings and who has not)

src/index.js
```diff
@@ -275,10 +269,10 @@ class UglifyJsPlugin {
           // Handling warnings
           if (warnings && warnings.length > 0) {
             warnings.forEach((warning) => {
-              const builtWarning = this.buildWarning(
+              const builtWarning = UglifyJsPlugin.buildWarning(
                 warning,
                 file,
-                inputSourceMap,
+                sourceMap,
```
evilebottnawi: Also `sourceMap` can be undefined here, because warning !== error (catch block)

lojjic:
> Right now your solution always try to parse source maps

That is not true, check the diff. Line 201 after the change.

src/index.js
evilebottnawi: /cc @lojjic yep, my mistake, but if no error(s) - no source map

evilebottnawi: again mistake, sorry, very tired

lojjic: I'm not following you. `sourceMap` can be `null`, if there is no valid `inputSourceMap`, but that's expected AFAICT.

lojjic: Ah, ok! No worries.

evilebottnawi:
Also i investigate why benchmark increase memory usage on `~25%`

src/index.js
```diff
@@ -187,11 +171,15 @@ class UglifyJsPlugin {
 
             tasks.push(task);
           } catch (error) {
+            let sourceMap = null;
```
michael-ciniawsky: `\n`

src/index.js
```diff
@@ -208,14 +196,20 @@ class UglifyJsPlugin {
           const { file, input, inputSourceMap, commentsFile } = tasks[index];
           const { error, map, code, warnings, extractedComments } = data;
 
+          // Create source map if needed
+          let sourceMap = null;
```
michael-ciniawsky: `\n`

lojjic:
Thanks for the bump. I understand the desire to fill in the test coverage for those scenarios you mentioned, if they were previously missing.

I've glanced at the tests and I'd definitely have to come up to speed with how the whole plugin works in context as well as how your tests are structured. Given my current time constraints I wouldn't expect that to happen in the next couple weeks.

evilebottnawi:
Close in favor https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/305

Thanks!

Issue #262, opened 2018-03-19
Question: Switch to maintained uglification module?


sirreal:
It appears that `uglify-es` will not be maintained. See https://github.com/mishoo/UglifyJS2/issues/2908#issuecomment-367686007 and https://github.com/mishoo/UglifyJS2/issues/3010#issuecomment-374147334.

There are [known issues](https://github.com/mishoo/UglifyJS2/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Abug+label%3Aharmony) which affect the behavior of code transformed by `uglify-es`. I recently came across a pesky production only bug (https://github.com/Automattic/wp-calypso/issues/21489) which was caused by an issue with `uglify-es` (https://github.com/mishoo/UglifyJS2/issues/3010).

I'd like to consider what can be done to ensure that this plugin reliably transforms code that will maintain the same behavior as pre-transform code. Some ideas:

- Switch to `uglify-js`
- Switch to another module like `babel-minify` (https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/264#issuecomment-374536624)
- Use a set of default options that disable uglification features known to be problematic (#264, #265)
- Encourage contributions to port `uglify-js` fixes to `uglify-es`

cc: @jsnajdr @ockham

evilebottnawi:
@sirreal thanks for issue, let's wait response from other contributors, i confused why `uglify-es` will not be maintained. How we can handle es5+ syntax

sirreal:
> How we can handle es5+ syntax 😕 ?

Not sure if this question is for me or other maintainers. I'll just reference #264 proposing a `babel-minify` as an alternative (https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/264#issuecomment-374536624).

estaub:
@evilebottnawi - see comments on mishoo/UglifyJS2/issues/3010, mishoo/UglifyJS2/issues/2842

...which suggests another option, a wee bit out of scope: use a forked, maintained uglify-es (a.k.a. "harmony" in the comments).

evilebottnawi:
@estaub do you have time to maintenance new package?

estaub:
@evilebottnawi I've not the skill, and if it wasn't obvious, I was being ironic in calling it "a wee bit out of scope"!  But this is a big enough problem that someone might pick it up and/or fund it.

[...]

PR #288, opened 2018-05-05
chore(package): switch to `@fabiosantoscode/uglify-es`

michael-ciniawsky:
### `Issues`

- #262

### `Notable Changes`

Switches from `uglify-es` (unmaintained)  to `@fabiosantoscode/uglify-es` (maintained)

### DIFF
--- a/src/uglify/versions.js
+++ b/src/uglify/versions.js
@@ -1,5 +1,4 @@
 export default {
-  uglify: require('uglify-es/package.json').version, // eslint-disable-line global-require
+  uglify: require('@fabiosantoscode/uglify-es/package.json').version, // eslint-disable-line global-require
   plugin: require('../../package.json').version, // eslint-disable-line global-require
 };
-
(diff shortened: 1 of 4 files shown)


src/uglify/versions.js
```diff
@@ -1,5 +1,4 @@
 export default {
-  uglify: require('uglify-es/package.json').version, // eslint-disable-line global-require
+  uglify: require('@fabiosantoscode/uglify-es/package.json').version, // eslint-disable-line global-require
   plugin: require('../../package.json').version, // eslint-disable-line global-require
 };
```
evilebottnawi: Lose newline in end of file?

src/uglify/versions.js
```diff
@@ -1,5 +1,4 @@
...
-
```
evilebottnawi: Here :smi

michael-ciniawsky: mom, I didn't intentionally touch it (maybe `eslint --fix` on `precommit`)

evilebottnawi: Very strange, let's return this newline

evilebottnawi:
Close in favor https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/293

Issue #303, opened 2018-05-27
[Feature Request] Allow cacheKey customization


Rohit-L:
Hello all,

Would it be possible to add an option to allow us to customize the cacheKey when caching?

Currently, this is the predefined cacheKey that is used.

```
            if (this.options.cache) {
              task.cacheKey = serialize({
                'uglify-es': versions.uglify,
                'uglifyjs-webpack-plugin': versions.plugin,
                'uglifyjs-webpack-plugin-options': this.options,
                path: compiler.outputPath ? `${compiler.outputPath}/${file}` : file,
                hash: crypto.createHash('md4').update(input).digest('hex'),
              });
            }
```

My use case is that on my production build machine, I will clone my repo into a temporary folder. The name of the temp folder is the git hash. My goal is to have the uglify cache reside outside this temp folder (which is doable as the cache option takes a string which is the cache directory).

My issue is that the cache doesn't help in this case as the cacheKey takes `compiler.outputPath` into consideration (the output path contains my temp folder name). Therefore, nothing will hit in the cache.

An option to allow me to customize the cacheKey that is used would let me define it so that `compiler.outputPath` is not used in the cacheKey.

evilebottnawi:
@Rohit-L 
1. Very very very weird use case. Looks you are trying to do something here, what should be do in another way. You should review you workflow with caches. It is just recommendation.
2. `additionalCacheKeys` - options, also please update docs and write tests on this (PR welcome).

For other developers: in near future we allow using `uglif-js` and other forks based on `uglify` have same api. For invalidate cache we should add option.

evilebottnawi:
Done in https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/320

PR #304, opened 2018-05-28
feature: add support for additional cache keys option

Rohit-L:
PR for issue #303 

This adds an option for user to pass additional keys to the default cache key generator for more customization, allowing the user to have more control, if needed, over what hits/misses in the cache.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ module.exports = {
 |**`include`**|`{RegExp\|Array<RegExp>}`|`undefined`|Files to `include`|
 |**`exclude`**|`{RegExp\|Array<RegExp>}`|`undefined`|Files to `exclude`|
 |**`cache`**|`{Boolean\|String}`|`false`|Enable file caching|
+|**`additionalCacheKeys`**|`{Object}`|`{}`|Add additional keys to the default cache key generator|
 |**`parallel`**|`{Boolean\|Number}`|`false`|Use multi-process parallel running to improve the build speed|
 |**`sourceMap`**|`{Boolean}`|`false`|Use source maps to map error message locations to modules (This slows down the compilation) ⚠️ **`cheap-source-map` options don't work with this plugin**|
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -112,6 +113,35 @@ Default path to cache directory: `node_modules/.cache/uglifyjs-webpack-plugin`.
...
+
+#### `{Object}`
+
+**webpack.config.js**
+```js
+[
+  new UglifyJsPlugin({
+    cache: true,
+    additionalCacheKeys: {
+      env: 'development',
```
evilebottnawi: Bad example, many people can use this from `README`, just use `myCustomVariable` name or something like

test/cache-options.test.js
```diff
@@ -436,4 +436,152 @@ describe('when options.cache', () => {
...
+
+    beforeEach(() => {
+      const pluginEnvironment = new PluginEnvironment();
+      const compilerEnv = pluginEnvironment.getEnvironmentStub();
+      compilerEnv.context = '';
+
+      const plugin = new UglifyJsPlugin({
+        cache: true,
+        additionalCacheKeys: {
+          env: 'development',
```
evilebottnawi: Please also use other variable

Rohit-L:
👍 Updated with suggestions from @evilebottnawi

evilebottnawi:
@ooflorent i am not fan also, but it should be documented because in some case it is only one variant to invalidate cache

shellscape:
In the past I've taken large, complex portions of a README and moved them out into separate `.md` files in a `./docs` directory. That might be considered to keep the README a little smaller. I would like to review this PR but I know practically nothing about how the plugin works, so I'm afraid that my review wouldn't be very useful.

evilebottnawi:
@shellscape i think we can do this in other PR, not here

Rohit-L:
@evilebottnawi @ooflorent what do you think about making `additionalCacheKeys` a function that returns the object of additional keys instead?

I just realized, for my use-case, I need access to `file` when defining my additional cache keys and this implementation wouldn't allow that.

I'm proposing something similar to how [cache-loader](https://github.com/webpack-contrib/cache-loader) defines their `cacheKey` option.

evilebottnawi:
@Rohit-L can you provide example?

evilebottnawi:
Close in favor https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/320

PR #306, opened 2018-05-31
fix: return missing important information

### LINKED ISSUE #281: Warning output missed important information
After upgrading the uglifyjs-webpack-plugin dependency in webpack I noticied that the warnings format changed:

test result

* Before: single warning for an asset, After: each warning separate
* Before: info about the asset, After: information missing
* Before: note about UglifyJS, After: no info

I propose to merge warnings by module and add a not about asset and uglifyjs-webpack-plugin. Example:

```
WARNING in ./index.js in bundle.js (uglifyjs-webpack-plugin)
6,0: Dropping side-effect-free statement
8,0: Dropping unused function someUnUsedFunction1
9,0: Dropping unused function someUnUsedFunction2
10,0: Dropping unused function someUnUsedFunction3
11,0: Dropping unused function someUnUsedFunction4
12,0: Dropping unused function someUnUsedFunction5

WARNING in ./a.js in bundle.js (uglifyjs-webpack-plugin)
3,0: Dropping unused function someRemoteUnUsedFunction1
4,0: Dropping unused function someRemoteUnUsedFunction2
5,0: Dropping unused function someRemoteUnUsedFunction3
6,0: Dropping unused function someRemoteUnUsedFunction4
7,0: Dropping unused function someRemoteUnUsedFunction5
```

evilebottnawi:
fixes #281  

> Before: single warning for an asset, After: each warning separate
Before: info about the asset, After: information missing
Before: note about UglifyJS, After: no info

1. Each warning separate, it is allow filtering `compilation.warning` for plugins and who run webpack using own script (i have many questions in gitter how it is better do and one warning on an assets is best solution)
2. Add information about the asset
3. Add information about `UglifyJS`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -106,7 +106,7 @@ class UglifyJsPlugin {
       }
     }
 
-    return warningMessage;
+    return `UglifyJs Plugin: ${warningMessage} in ${file}`;
   }
 
   apply(compiler) {
--- a/test/__snapshots__/all-options.test.js.snap
+++ b/test/__snapshots__/all-options.test.js.snap
@@ -57,6 +57,6 @@ exports[`when applied with all options matches snapshot: manifest.d6857f782c13a9
 
 exports[`when applied with all options matches snapshot: warnings 1`] = `
 Array [
-  "Dropping unused variable a [./test/fixtures/entry.js:4,0]",
+  "UglifyJs Plugin: Dropping unused variable a [./test/fixtures/entry.js:4,0] in main.0c220ec66316af2c1b24.js",
 ]
 `;


src/index.js
```diff
@@ -106,7 +106,7 @@ class UglifyJsPlugin {
       }
     }
 
-    return warningMessage;
+    return `｢UglifyJs｣ ${warningMessage} in ${file}`;
```
ooflorent: Not sure about the brackets. `validateOptions` does not add them.

evilebottnawi: @ooflorent just use `UglifyJs`?

ooflorent: Maybe `UglifyJs Plugin` just as here: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/6ab091801ae4e9a0ef59dcab950717a98923a037/src/index.js#L21

evilebottnawi:
@ooflorent done :+1:

PR #296, opened 2018-05-17
Update uglify-es dependency to use maintained terser module

fabiosantoscode:
I'm updating the `uglify-es` dependency to the new fork, which is now named `terser`

PR #308, opened 2018-06-09
fix: disable inline optimization by default

### LINKED ISSUE #264: Turn off inlining in optimization until UglifyJS issue is fixed.
This minimization bug has been open for almost two months: https://github.com/mishoo/UglifyJS2/issues/2842.  It has been referenced here elsewhere.  The bug occurs when UglifyJS attempts to inline a call to a function with a variable of the same name as a variable in the calling function.  Symptoms usually include "is not defined" or "assignment to constant variable".

In my experience, it's both easy to fall into and extremely time-consuming to diagnose.  

**I'd suggest turning `inline: false` in Webpack 4 until this is fixed.**  From the comments, it looks like UglifyJS2 is suffering from a lack of contributors with the skills and bandwidth needed to deal with things like this.

All of the Webpack 4 blogging has been about "Just let us do it for you."  IMHO, to fulfill the implied promise, I think you need to be reactive to component issues like this.

Related: `optimization.minimize` and `optimization.minimizer` (along with the rest of `optimization`) need documenting.  

For this particular issue, I'm using:
```
{
optimization:
            minimizer: [
               
                new UglifyJsPlugin({
                    parallel: true,  // Webpack default
                    cache: true      // Webpack default
                    uglifyOptions: {
                        /*
                            inlining is broken sometimes where inlined function uses the same variable name as inlining function.
                            See https://github.com/mishoo/UglifyJS2/issues/2842, https://github.com/mishoo/UglifyJS2/issues/2843
                         */
                        compress: { inline:false },
                    },
                })
            ],
}

```

---

This issue was moved from webpack/webpack#6798 by @evilebottnawi. Original issue was by @estaub.

### LINKED ISSUE #307: Webpack generated broken code in production mode
# Bug report

**What is the current behavior?**

Webpack generated broken code in production mode.

- The source code: [src/index.js](https://github.com/mysticatea/webpack-issue/blob/77dee08f91c4dabfdca0e7f9229148d688db63e6/src/index.js)
- The result code: [dist/main.js](https://github.com/mysticatea/webpack-issue/blob/77dee08f91c4dabfdca0e7f9229148d688db63e6/dist/main.js)

The result code contains a re-assignment to a `const` variable. It throws a fatal error in strict mode.

```js
const r=e[t];(r=r).fatal||r.ruleId||!r.message.includes("eslint-disable")||console.log(r)
```

It looks like a confusion about inlining local functions.

**If the current behavior is a bug, please provide the steps to reproduce.**

See https://github.com/mysticatea/webpack-issue

No loaders, plugins, and configurations.
It just does `webpack --mode production`.

**What is the expected behavior?**

It generates the code which can work.

**Other relevant information:**
webpack version:  4.11.1
Node.js version: 10.4.0
Operating System: Windows 10
Additional tools:

---

This issue was moved from webpack/webpack#7504 by @ooflorent. Original issue was by @mysticatea.

evilebottnawi:
fixes #307 fixes #264

Also we have problem with using option as object, they rewrited after first run.
Also improve readme
A lot of new tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,6 +42,9 @@ class UglifyJsPlugin {
       include,
       exclude,
       uglifyOptions: {
+        compress: {
+          inline: 1,
+        },
         output: {
           comments: extractComments ? false : /^\**!|@preserve|@license|@cc_on/,
         },
--- a/src/uglify/minify.js
+++ b/src/uglify/minify.js
@@ -21,9 +21,10 @@ const buildUglifyOptions = ({
 } = {}) => ({
   ecma,
   warnings,
-  parse,
-  compress,
-  mangle: mangle == null ? true : mangle,
+  parse: { ...parse },
+  compress: typeof compress === 'boolean' ? compress : { ...compress },
+  // eslint-disable-next-line no-nested-ternary
+  mangle: mangle == null ? true : typeof mangle === 'boolean' ? mangle : { ...mangle },
   output: {
     shebang: true,
     comments: false,
@@ -31,7 +32,7 @@ const buildUglifyOptions = ({
     semicolons: true,
     ...output,
   },
-  // Ignoring sourcemap from options
+  // Ignoring sourceMap from options
   sourceMap: null,
   toplevel,
   nameCache,
--- a/README.md
+++ b/README.md
@@ -164,8 +164,8 @@ Number of concurrent runs.
 |**`warnings`**|`{Boolean}`|`false`|Display Warnings|
 |**[`parse`](https://github.com/mishoo/UglifyJS2/tree/harmony#parse-options)**|`{Object}`|`{}`|Additional Parse Options|
 |**[`compress`](https://github.com/mishoo/UglifyJS2/tree/harmony#compress-options)**|`{Boolean\|Object}`|`true`|Additional Compress Options|
-|**[`mangle`](https://github.com/mishoo/UglifyJS2/tree/harmony#mangle-options)**|`{Boolean\|Object}`|`true`|Enable Name Mangling (See [Mangle Properties](https://github.com/mishoo/UglifyJS2/tree/harmony#mangle-properties-options) for advanced setups, use with ⚠️)|
-|**[`output`](https://github.com/mishoo/UglifyJS2/tree/harmony#output-options)**|`{Object}`|`{}`|Additional Output Options (The defaults are optimized for best compression)|
+|**[`mangle`](https://github.com/mishoo/UglifyJS2/tree/harmony#mangle-options)**|`{Boolean\|Object}`|`{inline: false}`|Enable Name Mangling (See [Mangle Properties](https://github.com/mishoo/UglifyJS2/tree/harmony#mangle-properties-options) for advanced setups, use with ⚠️)|
+|**[`output`](https://github.com/mishoo/UglifyJS2/tree/harmony#output-options)**|`{Object}`|`{comments: extractComments ? false : /^\**!|@preserve|@license|@cc_on/,}`|Additional Output Options (The defaults are optimized for best compression)|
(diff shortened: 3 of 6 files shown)


src/uglify/minify.js
```diff
@@ -21,17 +21,23 @@ const buildUglifyOptions = ({
...
+  },
   output: {
     shebang: true,
     comments: false,
     beautify: false,
     semicolons: true,
     ...output,
   },
-  // Ignoring sourcemap from options
+  // Ignoring sourceMap from options
```
evilebottnawi: Typo in option name

src/uglify/minify.js
```diff
@@ -21,17 +21,23 @@ const buildUglifyOptions = ({
...
   ecma,
   warnings,
-  parse,
-  compress,
-  mangle: mangle == null ? true : mangle,
+  parse: {
+    ...parse,
+  },
+  compress: {
+    ...compress,
```
kzc: this won't handle `compress: false`

src/uglify/minify.js
```diff
@@ -21,17 +21,23 @@ const buildUglifyOptions = ({
...
+  },
+  mangle: mangle == null ? true : {
+    ...mangle,
```
kzc: this won't handle `mangle: false`

src/index.js
```diff
@@ -42,6 +42,9 @@ class UglifyJsPlugin {
       include,
       exclude,
       uglifyOptions: {
+        compress: {
+          inline: false,
```
kzc: `inline: 1` is known to work in `uglify-es@3.3.9`.

```
$ echo '!function(){ console.log(42); }();' | bin/uglifyjs -c inline=false
!function(){console.log(42)}();
```
```
$ echo '!function(){ console.log(42); }();' | bin/uglifyjs -c inline=1
console.log(42);
```

kzc:
If this plugin is upgraded to use `terser` in the future (PR #296), then the `inline` override can be reverted.

src/uglify/minify.js
kzc: nor will it handle `compress: true`

src/uglify/minify.js
kzc: nor will it handle `mangle: true`

evilebottnawi:
@kzc we can not merge use `terser` as `patch/minor` release, it is require major release. 
How we can solve this:
1. Disable `inline` for `1` version (patch release).
2. Allow people using own `cache` keys https://github.com/webpack-contrib/uglifyjs-webpack-plugin/pull/304 (need implement `function` support) (minor release).
3. Implement feature using own `minify` function (minor release).
4. Documented how plugin can setup using `uglify-js` and `terser` (patch release).
5. Switch on `terser` for next major version (major release).

kzc:
> we can not merge use terser as patch/minor release, it is require major release

I disagree. `terser` is a drop in replacement for `uglify-es`. It's a minor change.

Disabling `inline` in `uglify-es` only addresses a couple of bugs. And it could introduce other bugs in this uglify wrapper plugin as per my PR review comments above.

evilebottnawi:
@kzc other developers disagree with
> I disagree. terser is a drop in replacement for uglify-es. It's a minor change.

All I'm trying to do is find the best solution for everyone and I'm tired of being between these disputes. You suggest using `minor` release (only relying on your words) it's too dangerous and better use major release, it is allow people safely upgrade uglify plugin (If the update introduces new bugs they can roll back without fixing the version in `package.json`).

kzc:
I've been involved with the development of both `uglify-es` and `terser` since their inception. Terser is just a minor fork of `uglify-es` with a name change with dozens of bug fixes. But if you're more comfortable with a major for that, so be it. `webpack@4` users would have to explicitly load the new major plugin to get the bug fixes.

evilebottnawi:
@kzc After release major version with `terser` we wait some times to ensure all works as expected, when bump this in webpack@4

shellscape:
LGTM

PR #325, opened 2018-07-03
feat: support `minify` function option

evilebottnawi:
:star: 

One question: what do with extracting comment for custom minify function, if people use `uglify-js`/`terser` our logic https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/src/uglify/minify.js#L45 is work good, maybe we should export this function to avoid copy/paste this code

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ module.exports = {
 |**`cacheKeys`**|`{Function(defaultCacheKeys, file) -> {Object}}`|`defaultCacheKeys => defaultCacheKeys`|Allows you to override default cache keys|
 |**`parallel`**|`{Boolean\|Number}`|`false`|Use multi-process parallel running to improve the build speed|
 |**`sourceMap`**|`{Boolean}`|`false`|Use source maps to map error message locations to modules (This slows down the compilation) ⚠️ **`cheap-source-map` options don't work with this plugin**|
+|**`minify`**|`{Function}`|`undefined`|Allows you to override default minify function|
 |**`uglifyOptions`**|`{Object}`|[`{...defaults}`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/master#uglifyoptions)|`uglify` [Options](https://github.com/mishoo/UglifyJS2/tree/harmony#minify-options)|
 |**`extractComments`**|`{Boolean\|RegExp\|Function<(node, comment) -> {Boolean\|Object}>}`|`false`|Whether comments shall be extracted to a separate file, (see [details](https://github.com/webpack/webpack/commit/71933e979e51c533b432658d5e37917f9e71595a) (`webpack >= 2.3.0`)|
 |**`warningsFilter`**|`{Function(source) -> {Boolean}}`|`() => true`|Allow to filter uglify warnings|
@@ -88,6 +89,8 @@ module.exports = {
 ```
 
 ### `cache`
+
+If you use your own `minify` function please read the `minify` section for cache invalidation correctly.
 
 #### `{Boolean}`
 
@@ -178,6 +181,8 @@ Number of concurrent runs.
 
 ### `sourceMap`
 
+If you use your own `minify` function please read the `minify` section for handling source maps correctly.
+
 **webpack.config.js**
 ```js
 [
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -89,6 +90,8 @@ module.exports = {
 
 ### `cache`
 
+If you use custom `minify` function please read `minify` option section for cache invalidation correctly.
```
ev1stensberg: a custom

README.md
```diff
@@ -178,6 +181,8 @@ Number of concurrent runs.
 
 ### `sourceMap`
 
+If you use custom `minify` function please read `minify` option section for handle source maps correctly.
```
ev1stensberg: `If you use custom `minify` function please read `minify` option section for handle source maps correctly.` -> `If you use your own `minify` function please read the `minify` section for handling source maps correctly.`

README.md
```diff
@@ -189,6 +194,102 @@ Number of concurrent runs.
...
+    // cache: true,
+    // cacheKeys(defaultCacheKeys) {
+    //   return Object.assign(
+    //     {},
+    //     defaultCacheKeys,
+    //     { 'uglify-js': require('uglify-js/package.json').version },
+    //   );
+    // },
+    minify(file, sourceMap) {
+      // Look https://github.com/mishoo/UglifyJS2#minify-options
```
ev1stensberg: Remove Look, a link is enough perhaps

README.md
```diff
@@ -189,6 +194,102 @@ Number of concurrent runs.
...
+    // cache: true,
+    // cacheKeys(defaultCacheKeys) {
+    //   return Object.assign(
+    //     {},
+    //     defaultCacheKeys,
+    //     { terser: require('terser/package.json').version },
+    //   );
+    // },
+    minify(file, sourceMap) {
+      // Look https://github.com/fabiosantoscode/terser#minify-options
```
ev1stensberg: Strip `Look`

test/minify.test.js
```diff
@@ -0,0 +1,170 @@
...
+        });
+      },
+    }).apply(compiler);
+
+    return compile(compiler)
+      .then((stats) => {
+        const errors = stats.compilation.errors.map(cleanErrorStack);
+        const warnings = stats.compilation.warnings.map(cleanErrorStack);
+
+        expect(errors).toMatchSnapshot('errors');
```
ev1stensberg: Don't think this is needed, jest will resolve it

test/minify.test.js
```diff
@@ -0,0 +1,170 @@
...
+        path: `${__dirname}/dist-terser`,
+        filename: '[name].js',
+        chunkFilename: '[id].[name].js',
+      },
+    });
+
+    function removeAbsoluteSourceMapSources(source) {
+      if (source.map && source.map.sources) {
+        // eslint-disable-next-line no-param-reassign
+        source.map.sources = source.map.sources.map(sourceFromMap => path.relative(process.cwd(), sourceFromMap));
```
ev1stensberg: any way to copy and reassign instead of directly mutating here? Perhaps extracting into a variable and reassigning would be better https://stackoverflow.com/questions/34716651/js-array-prototype-map-happens-to-be-mutable

test/minify.test.js
```diff
@@ -0,0 +1,170 @@
...
+      .then((stats) => {
+        const errors = stats.compilation.errors.map(cleanErrorStack);
+        const warnings = stats.compilation.warnings.map(cleanErrorStack);
+
+        expect(errors).toMatchSnapshot('errors');
+        expect(warnings).toMatchSnapshot('warnings');
+
+        for (const file in stats.compilation.assets) {
+          if (Object.prototype.hasOwnProperty.call(stats.compilation.assets, file)) {
+            expect(removeAbsoluteSourceMapSources(stats.compilation.assets[file].sourceAndMap())).toMatchSnapshot(file);
```
ev1stensberg: assertion is a bit long, maybe a variable to hold the assertion value would be good

test/minify.test.js
evilebottnawi: @ev1stensberg will be done in other PR (refactor tests) :+1:

test/minify.test.js
evilebottnawi: @ev1stensberg big in snapshots? I think it is normal, it is allow to catch some strange problem with source map and uglification

ev1stensberg: Alright. Thought it would be better to test variables containing the values, but that might be hard with compilation instances

README.md
```diff
@@ -189,6 +194,102 @@ Number of concurrent runs.
...
+         { /* Your options for minification */ },
+       );
+
+       return { error, map, code, warnings, extractedComments };
+      }
+  })
+]
+```
+
+By default plugin use `uglify-es` package.
```
ooflorent: _uses_?

evilebottnawi: @ooflorent same people want to use own minification (forks) and without this low level function it is impossible, creating new plugin for this situation is not good solution, because it is require copy/paste 99% code of plugin, also we other fix small problems, this will require people to constantly monitor this and cause a lot of inconvenience. Also it is simple way to test alpha/beta/rc/etc versions.

test/minify.test.js
evilebottnawi: @ev1stensberg why? It is increase lines of code and decrease perf, but we just need remove absolute path. I don't think it is critical here.

kzc:
Something to consider... if the default `minify` function is changed, the version remains `"uglify-es":"3.3.9"` in the `uglifyjs-webpack-plugin` cache key:

https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/0174605fdf7b0eb8dc3b18aaec262d1e757e6863/src/index.js#L177-L179

and the cached assets will be invalid.

kzc:
> uglify-es is not updated long time, cache will works as expected

Until the module with the minify replacement function is upgraded - terser, uglify-js, or whatever. You might mention in the docs if this minify field is overriden or dependencies are updated then the user should delete their build cache. You could also elect to allow the user to provide a custom version string to workaround this issue.

evilebottnawi:
Also we provide way to invalidate cache https://github.com/webpack-contrib/uglifyjs-webpack-plugin#terser

PR #333, opened 2018-07-27
Refactor stuff

evilebottnawi:
Briefly:
1. Move some utils to class (we don't need extract require for some stuff).
2. Merge some identical tests.
3. Add test for validation.
4. Rename all test (example `sourceMap.test.js` -> `sourceMap-option.test.js`)


src/index.js
```diff
@@ -56,8 +54,18 @@ class UglifyJsPlugin {
     };
   }
 
+  static isSourceMap(input) {
```
ooflorent: I prefer the function as a helper. If you expose it it becomes a public API

evilebottnawi:
Merge as is, feel free to feedback

PR #338, opened 2018-08-08
fix(uglify/Runner): prevent errors on limited shells where `os.cpus` returns `undefined`

vielhuber:

### DIFF
--- a/src/uglify/Runner.js
+++ b/src/uglify/Runner.js
@@ -16,7 +16,10 @@ export default class Runner {
   constructor(options = {}) {
     const { cache, parallel } = options;
     this.cacheDir = cache === true ? findCacheDir({ name: 'uglifyjs-webpack-plugin' }) : cache;
-    this.maxConcurrentWorkers = parallel === true ? os.cpus().length - 1 : Math.min(Number(parallel) || 0, os.cpus().length - 1);
+    // https://github.com/nodejs/node/issues/19022
+    // in some cases cpus() returns undefined
+    const cpusLength = (os.cpus() ? os.cpus().length : 1);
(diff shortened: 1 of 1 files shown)


src/uglify/Runner.js
```diff
@@ -16,7 +16,10 @@ export default class Runner {
   constructor(options = {}) {
     const { cache, parallel } = options;
     this.cacheDir = cache === true ? findCacheDir({ name: 'uglifyjs-webpack-plugin' }) : cache;
-    this.maxConcurrentWorkers = parallel === true ? os.cpus().length - 1 : Math.min(Number(parallel) || 0, os.cpus().length - 1);
+    // https://github.com/nodejs/node/issues/19022
+    // in some cases cpus() returns undefined
+    const cpusLength = (os.cpus() ? os.cpus().length : 1);
```
evilebottnawi: Let's rewrite to `const cpusLength = (os.cpus() || { length: 1 }).length` as said https://github.com/nodejs/node/issues/19022#issuecomment-374275868

vielhuber: Sorry, I don't know how this is done. Can you help out?

evilebottnawi: @vielhuber change code in you branch, commit changes and push with force. Sorry i don't have time to learn, better read book about `git`. You can create new PR if you don't know how do this.

vielhuber: Done.

evilebottnawi:
@michael-ciniawsky some people use `chroot` when using `linux`, problem really exists in node and it is break plugin (i tested with locally and confirm problem exists). Explanation is not required.

PR #346, opened 2018-08-25
feat: support passing context object to minify function option

clydin:
This allows the minify function to be passed contextual information providing a method for minified settings to be calculated based on external factors when the `parallel` option is enabled.

Hypothetical Example:

```
const minifyContext = {
  noMangle: true
};

const plugin = new UglifyJsPlugin({
  parallel: true,
  minifyContext,
  minify(file, sourceMap, context) {
    // eslint-disable-next-line global-require
    return require('terser').minify(file, {
      mangle: !context.noMangle,
    });
  },
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ module.exports = {
 |**`parallel`**|`{Boolean\|Number}`|`false`|Use multi-process parallel running to improve the build speed|
 |**`sourceMap`**|`{Boolean}`|`false`|Use source maps to map error message locations to modules (This slows down the compilation) ⚠️ **`cheap-source-map` options don't work with this plugin**|
 |**`minify`**|`{Function}`|`undefined`|Allows you to override default minify function|
+|**`minifyContext`**|{Object}|`undefined`|Allows passing contextual information to the minify function.  Useful in combination with `parallel`.  The object must be serializable.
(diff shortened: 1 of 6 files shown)


clydin:
The parallel option is the reason for this change.  Since the minify function is directly serialized it cannot act as a closure and therefore prevents data from being closed around the function.  This provides a method to still allow data to be used within the function without the need for complex shared memory implementations.

As implemented, the minify function option is essentially useless outside of trivial applications.  Note also that the terser usage was purely hypothetical and any other minifier would also suffer from this problem.

This allows static pre-calculated options to be used within the minify function.  This is a requirement for more complex webpack configuration setups that adjust settings based on the execution context and environment (e.g., environment variables, passed command line options, wrapping the config within a larger system, etc.).

evilebottnawi:
@clydin for these purpose you should use `SharedMemory` (it is allow transfer objects/function/string and etc to other process). There are several implementations on nodejs, feel free to experiment with this.

clydin:
It is your call but I respectively disagree.  Using share memory is the equivalent of using a sledgehammer on a screw when a screwdriver would suffice.  In general, shared memory should be avoided at all costs due to its complexity, data integrity, and most importantly security related concerns.  Except for low-level, high-performance IPC related scenarios, there are almost always better options.

However, we are moving away from this plugin so there is no issue with closing.  It was useful for testing out other potential options but we can create small standalone plugins for that purpose if needed.

Thank you for your time and consideration.

PR #356, opened 2018-09-17
fix: ping uglify-js to an older version

### LINKED ISSUE #355: UfligyJS un-necessarily pinned at the latest and buggy release
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/package.json#L50

pins uglify to 3.4.9

But there are some bad uglify-js bugs with 3.4.9:

https://github.com/mishoo/UglifyJS2/issues/3245
https://github.com/mishoo/UglifyJS2/issues/3257
https://github.com/mishoo/UglifyJS2/issues/3260

and now as a consumer of this plugin I cannot specify the uglify-js version in my parent package.json in order to use a non-buggy release.

lukeapage:
This allows users of the plugin to pin to older versions to avoid bugs.
3.1.4 picked purely because its in the package lock file as what handlebars pins to.
Fixes #355

This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

See bug #355

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "schema-utils": "^1.0.0",
     "serialize-javascript": "^1.4.0",
     "source-map": "^0.6.1",
-    "uglify-js": "^3.4.9",
+    "uglify-js": "^3.1.4",
     "webpack-sources": "^1.1.0",
     "worker-farm": "^1.5.2"
   },


package.json
```diff
@@ -47,7 +47,7 @@
     "schema-utils": "^1.0.0",
     "serialize-javascript": "^1.4.0",
     "source-map": "^0.6.1",
-    "uglify-js": "^3.4.9",
+    "uglify-js": "^3.1.4",
```
evilebottnawi: Why not just `^3.0.0`?

Issue #21, opened 2017-03-23
Expose chunk to `options.<test|include|exclude>`


japgolly:
The `test`, `include` and `exclude` options only inspect the output filename, which has no value when the names are like `[hash].[ext]`. It would be very helpful to expose the associated chunk info as well.

hulkish:
@japgolly How about this for usage:
````js
new UglifyJsPlugin ({ 
  // same as current behavior, would convert old syntax to this for legacy support.
  assets: {
    test: /foo.js/,
    include: /bar.js/,
    exclude: require.resolve('../some/path'),
  },
  modules: {
    include: require.resolve('../src/dont-uglify-me-specifically')
  },
});
```

I believe this offers basically the finest grain of control.

FranckFreiburger:
I have created a loader for [javascript-obfuscator](https://github.com/javascript-obfuscator/javascript-obfuscator), and I would like to avoid uglifyjs-webpack-plugin to process resources that pass to that loader (because obfuscated code must not be uglified). 

I would like to use something like:
```
  modules(module) (
    if ( module.loaders.find(loader => .... 'javascript-obfuscator-loader' ....) !== null ) {
      return false
    }
    return true;
  ),
```

Friksel:
+1 
A way to exclude libs from uglifying/mangling would be very welcome!

evilebottnawi:
Will be shipped in near future

PR #369, opened 2018-10-20
feat(options): add chunkTest option

WearyMonkey:
Adds a chunkTest option to filter chunks.

This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The existing test/include/exclude options are not useful when the filenames are generated at buildtime, e.g. `[hash].[ext]`.

### Additional Info

Related to: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/21


WearyMonkey:
@evilebottnawi @michael-ciniawsky any thoughts?

evilebottnawi:
@WearyMonkey Please read my review above and fix problem

WearyMonkey:
@evilebottnawi I'm only seeing:

> Better name for option/function chunkFilter

I renamed `chunkTest` to `chunkFilter` as you suggested. 

> also it should be before .filter(ModuleFilenameHelpers.matchObject.bind(null, this.options)), not first

Didn't understand this sorry. Can you provide a code snippet? I thought the chunk filter should go before the chunk file accumulator.

Am I missing something else? I think you commented during the github downtime so maybe something was lost.

WearyMonkey:
@evilebottnawi as far as I can tell I've responded to all your feedback. Am I missing something?

evilebottnawi:
Need fix test on windows

evilebottnawi:
@WearyMonkey will be great :+1:

WearyMonkey:
@evilebottnawi I ran `npm test -- -u` on windows, and seems to of touched all the snapshots, and the CI is still failing.

Not sure where to go from here?

PR #391, opened 2019-01-15
feat: don't compute input source maps if source file hasn't changed

kevinbarabash:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The plugin does a lot of work computing input source maps when it isn't necessary, i.e. the cache key hasn't changed and we already have a source map in our cache for the key.

### Breaking Changes

This shouldn't be a breaking change, but I'm not super familiar with all of the possible configurations for this plugin.

### Additional Info

This change looks a lot bigger than it actually is.  Because of the `cacache.get.info()` call that's introduced index, I had to wrap a big chunk of `optimizeFn` in a `Promise.all()` call because `cacache.get.info()` returns a promise and before this everything in `optimizeFn` was synchronous.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,6 +11,8 @@ import ModuleFilenameHelpers from 'webpack/lib/ModuleFilenameHelpers';
 import validateOptions from 'schema-utils';
 import serialize from 'serialize-javascript';
 import uglifyJsPackageJson from 'uglify-js/package.json';
+import cacache from 'cacache';
+import findCacheDir from 'find-cache-dir';
 
 import schema from './options.json';
 import TaskRunner from './TaskRunner';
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -170,224 +172,232 @@ class UglifyJsPlugin {
...
+                input,
+                inputSourceMap,
+                commentsFile,
+                extractComments: this.options.extractComments,
+                uglifyOptions: this.options.uglifyOptions,
+                minify: this.options.minify,
+                cacheKey,
               };
 
-              task.cacheKeys = this.options.cacheKeys(defaultCacheKeys, file);
```
kevinbarabash: I moved the computation of the cache keep up so that we can check whether the key exists in the cache before computing the input source map.

src/index.js
```diff
@@ -170,224 +172,232 @@ class UglifyJsPlugin {
...
+                    .update(input)
+                    .digest('hex'),
+                };
+
+                cacheKey = serialize(defaultCacheKeys);
+                cacheHit = !!(await cacache.get.info(cacheDir, cacheKey));
+              }
 
-            if (this.options.sourceMap && asset.sourceAndMap) {
-              const { source, map } = asset.sourceAndMap();
```
kevinbarabash: Apparently `sourceAndMap` is supposed to be faster than calling `asset.source()` and `asset.map()` separately, but not having to call `asset.map()` if we have the output source map cached is even faster. :)

src/index.js
```diff
@@ -170,224 +172,232 @@ class UglifyJsPlugin {
...
+                  // eslint-disable-next-line global-require
+                  'uglifyjs-webpack-plugin': require('../package.json').version,
+                  'uglifyjs-webpack-plugin-options': this.options,
+                  hash: crypto
+                    .createHash('md4')
+                    .update(input)
+                    .digest('hex'),
+                };
+
+                cacheKey = serialize(defaultCacheKeys);
```
kevinbarabash: We do the serialization of the cache keys once and pass that to the TaskRunner so we don't have to do it again.

src/index.js
```diff
@@ -170,224 +172,232 @@ class UglifyJsPlugin {
...
-              )
-            );
-          }
-        });
-
-      taskRunner.run(tasks, (tasksError, results) => {
-        if (tasksError) {
-          compilation.errors.push(tasksError);
+      promise.then(() => {
+        taskRunner.run(tasks, (tasksError, results) => {
```
kevinbarabash: The callback to `taskRunner` is unchanged, I've only wrapped it in a `.then()` now that the computation of `tasks` is async.

kevinbarabash:
whoops... I meant to make this a PR against my fork.

